Issue #625, opened 2026-04-02
feat: AnalyzerRule + TableProvider for bare-name SQL over typed columns


strawgate:
## Summary

Replace the dead 772-line SQL text rewriter (`rewriter.rs`) with a DataFusion `AnalyzerRule` + custom `TableProvider` that presents bare column names while preserving per-document type fidelity.

## Background

The scanner produces suffixed columns when a field has multiple types in a batch (`status_int` + `status_str`). The current SQL rewriter attempted to let users write bare names (`WHERE status > 400`) by rewriting SQL text, but it was never wired in (#602), could never handle all SQL constructs, and is the root cause of 8+ bugs (#415, #429, #430, #444).

Research into DataFusion's extension points found that `AnalyzerRule` + `ExprPlanner` + custom `TableProvider` solve this problem correctly at the plan level.

### Custom TableProvider (bare-name presentation)

- `schema()` returns bare column names: `status` (Utf8), `level` (Utf8)
- `get_logical_plan()` returns a plan with COALESCE expressions:
  ```sql
  SELECT COALESCE(CAST(status_int AS VARCHAR), status_str) AS status, ...
  FROM _raw_logs
  ```
- `SELECT *` shows bare names. Users never see `_int`/`_str`/`_float` suffixes.
- For clean-schema batches (OTLP, CSV, consistent JSON), no suffixed columns exist — the TableProvider passes through bare names directly.

### AnalyzerRule (type-aware column routing)

Registered via `ctx.add_analyzer_rule()`. Runs before type coercion. Walks the expression tree:

- `status > 400` (bare name vs numeric literal) → rewrites to `status_int > 400`
- `status = 'ERROR'` (bare name vs string literal) → rewrites to `status_str = 'ERROR'`
- `SELECT status` (no comparison context) → resolves to COALESCE string via the view
- Works for all SQL constructs automatically because it operates on the typed expression tree, not SQL text

### ExprPlanner (optional, earlier hook)

`ExprPlanner::plan_binary_op` can intercept comparisons even earlier, during SQL→LogicalPlan conversion. May be simpler than a full AnalyzerRule for the comparison case.

### Storage (unchanged)

Internal Arrow columns keep the suffix convention:
- Single type in batch: bare name, native DataType (`status` as Int64)
- Type conflict in batch: suffixed names (`status_int` + `status_str`)

The AnalyzerRule adapts to both cases — when columns are already bare, no rewriting needed.

### Schema stability

[...]

PR #684, opened 2026-04-02
feat: suffix column names only on type conflict, delete dead rewriter (#445)

### LINKED ISSUE #445: Column type design: suffix only on conflict, bare names by default
## Design

**Bare column names by default. Suffixed columns only when a field has multiple types within a batch.**

See `dev-docs/research/type-suffix-redesign.md` for the full design.

### Core requirements

1. **Round-trip type fidelity** — int in, int out. `SELECT *` preserves per-document types.
2. **Clean schemas stay clean** — OTLP, CSV, consistent JSON get bare column names. No suffixes, no views, no overhead.
3. **Mixed types handled gracefully** — JSON with `status: 200` and `status: "OK"` in the same batch produces `status_int` + `status_str`.

### How it works

| Input | Type conflicts? | Column names |
|---|---|---|
| OTLP / Arrow IPC | Never | Bare: `status` (Int64), `level` (Utf8) |
| CSV / raw | Never | Bare: `status` (Utf8), `level` (Utf8) |
| JSON (consistent) | No | Bare: `status` (Int64), `level` (Utf8) |
| JSON (mixed) | Yes | `status_int` + `status_str`, `level` (Utf8) |

### SQL interface

- Bare names work for clean schemas: `WHERE status > 400`
- Suffixed names for type-conflict fields: `WHERE status_int > 400`
- View layer maps bare names to string representation for conflict fields
- `SELECT *` always round-trips correctly

### Implementation phases

1. **Delete dead code** — `rewriter.rs` (772 lines, dead code per #602), `strip_type_suffix()`
2. **Suffix only on conflict** — builder change: bare names when single type
3. **Output grouping** — `build_col_infos()` uses DataType, not suffix parsing
4. **Schema stability** — pad SQL-referenced columns with nulls
5. **View layer** — auto-generate view when suffixed columns exist

### Bugs fixed

- #415 — SQL rewriter missing arms (rewriter deleted)
- #429 — EXCEPT clause not rewritten (rewriter deleted)
- #430 — SQL computed columns output empty strings (DataType dispatch)
- #404 — i64 overflow mixed int/float
- #442 — JSON null values silently dropped
- #444 — float 1.0 → integer 1
- #531 — schema evolution breaks rewriter (rewriter deleted)
- #410 — double-escaping in output (DataType dispatch)

strawgate:
## Summary

Implements the \"suffix only on conflict\" column naming design from #445.

- **Single-type fields** → bare column name with native type: `status` (Int64), `level` (Utf8)
- **Type-conflict fields** → suffixed names: `status_int` (Int64) + `status_str` (Utf8View)
- **Deletes `rewriter.rs`** — 775 lines of SQL text rewriter that was never wired into the pipeline (#602)

The output layer already dispatched on Arrow `DataType` rather than column name suffix (commit 0321b18), so serialization is unaffected.

## Changes

| File | Change |
|---|---|
| `streaming_builder.rs`, `storage_builder.rs` | Conflict detection in `finish_batch()` — bare name when single-type, suffixed on conflict |
| `rewriter.rs` | **Deleted** (775 lines, dead code) |
| `json_extract.rs` | `suffix_order` tries bare name as fallback; non-numeric columns return null instead of coercing `"200"` → `200` |
| All test files | Bare names for single-type fields; suffixed names preserved where genuine conflicts exist |
| `ci.yml` | `RUSTC_WRAPPER: ""` so `cargo test` works in CI without sccache |
| `type-suffix-redesign.md` | Updated VIEW reference → AnalyzerRule approach (see #625) |

## Test plan

- [ ] All existing tests pass with bare names for single-type fields
- [ ] Type-conflict tests still produce `_int`/`_str`/`_float` suffixes
- [ ] `json_extract` UDF: `json_int` on a quoted string returns null (not the coerced int value)
- [ ] CI passes

## Closes

Partially closes #445 (steps 1-3 complete; schema padding and AnalyzerRule tracked in #625).

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,836 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+inheritance: true
+reviews:
+  pre_merge_checks:
+    docstrings:
+      mode: 'off'
+    title:
+      requirements: >-
+        Title must follow conventional commit format under 70 characters:
+        type(optional-scope): concise description (#issue-number).
+        Allowed types: feat, fix, refactor, docs, bench, test, chore.
+        Scope is the crate name when the change is crate-specific
+        (e.g., feat(logfwd-core): add varint overflow proof (#123)).
+        The description must name the observable behavior change, not
+        the mechanism. Reject vague titles like "update scanner" or
+        "fix bug". Require "fix: CRI aggregator drops final partial line
+        on EOF (#275)" — specific, behavioral, linked. The issue number
+        is required for feat and fix commits; optional for refactor, docs,
+        bench, test, chore. PR titles must match the squash commit message
+        that will land on master.
+    description:
+      mode: 'off'
+    custom_checks:
+      - name: High-Quality Rust Practices
+        mode: error
+        instructions: >-
+          You are a senior Rust engineer reviewing a pull request against logfwd,
+          a formally verified, ultra-performance log forwarder targeting 1.7 million
+          lines/second on a single ARM64 core. You enforce the project's Rust idioms
+          ruthlessly. Fail this check if any of the following rules are violated anywhere
+          in changed code.
+
+
+          PANICS AND ERROR HANDLING.
+          Reject any .unwrap() call in production paths (non-test, non-kani code).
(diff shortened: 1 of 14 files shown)


strawgate:
@coderabbitai please do a very thorough review

Issue #1224, opened 2026-04-05
fix(ci): checks:write permission granted workflow-wide instead of scoped to lint job


strawgate:
PR #1160 added `checks: write` at the top-level `permissions:` block of `ci.yml`, granting expanded `GITHUB_TOKEN` privileges to every job in the workflow. Only the `lint` job's `rustsec/audit-check` step needs this permission.

Fix: move `checks: write` into `jobs.lint.permissions` and keep the top-level permissions read-only.

Issue #1219, opened 2026-04-05
fix(ci): fuzz-short nightly job missing sccache setup — will fail with wrapper not found


strawgate:
`.cargo/config.toml` sets `rustc-wrapper = "sccache"` workspace-wide. PR #1112 added sccache to some nightly jobs but left it out of `fuzz-short`. Every nightly fuzz run fails because `sccache` is not on PATH.

Fix: add `mozilla-actions/sccache-action` before the fuzz step, or override `RUSTC_WRAPPER: ""` for that job.

PR #1240, opened 2026-04-05
fix(ci): scope permissions, fix exit-code masking, add sccache to fuzz

### LINKED ISSUE #1218: fix(ci): nightly turmoil seed sweep masks cargo test exit code — failures always pass
In the nightly CI workflow (introduced #1098, extended #1134):

```bash
cargo test ... 2>&1 | tail -3
if [ $? -ne 0 ]; then ...
```

`tail` always exits 0, so `$?` is never non-zero. The `FAILED` counter never increments. Every nightly run reports success even when turmoil seeds find failures.

Fix: use `set -o pipefail` at the top of the shell step, or capture the exit code before piping:
```bash
cargo test ... > out.txt 2>&1
result=$?
tail -3 out.txt
if [ $result -ne 0 ]; then ...
```

Also: `timeout 300 cargo fuzz ... || true` forces success even when a fuzz target finds a crash. The `|| true` should be removed or the fuzz exit code handled explicitly.

strawgate:
Fixes #1218, #1219, #1224

- Scope checks:write to lint job only (#1224)
- Fix turmoil seed loop exit-code masking via pipe (#1218)
- Add sccache-action to fuzz job (#1219)

> [!NOTE]
> ### Fix exit-code masking in turmoil seed sweep and add sccache to fuzz job
> - Scopes `checks: write` permission to the lint job only, removing it from the workflow-level permissions in [ci.yml](.github/workflows/ci.yml).
> - Fixes exit-code masking in the turmoil seed sweep: test output is now captured to a file, the binary's exit code is preserved, and failure is triggered on non-zero exit or missing `ok` in the last output line.
> - Adds `sccache` to the fuzz job and removes the `timeout 1800 || true` wrapper, so fuzz targets now fail the job on error instead of being silently ignored; time limiting relies on cargo-fuzz's `-max_total_time=1800`.
> - Behavioral Change: fuzz failures that were previously suppressed will now fail the CI job.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b59a540.</sup>
> 
>

### DIFF
--- a/.github/workflows/nightly-testing.yml
+++ b/.github/workflows/nightly-testing.yml
@@ -80,5 +80,5 @@ jobs:
           cd crates/logfwd-core
           for target in $(cargo fuzz list 2>/dev/null); do
             echo "=== Fuzzing $target for 30 minutes ==="
-            timeout 1800 cargo fuzz run "$target" -- -max_total_time=1800
+            cargo fuzz run "$target" -- -max_total_time=1800
           done


strawgate:
Addressed MAJOR reviewer feedback:

**Fixed: outer `timeout` race condition** — Removed `timeout 1800` wrapper. `cargo fuzz run` exits 0 when `-max_total_time` is reached, but `timeout` fires simultaneously and returns exit code 124, causing nightly CI to report failure on a successful fuzz run. The fuzzer controls its own termination via `-max_total_time=1800`.

PR #1242, opened 2026-04-05
fix(logfwd-output): OTLP sink retry and gRPC Content-Encoding fixes

strawgate:
Fix the OTLP sink's double-retry bug and incorrect gRPC Content-Encoding header. The worker pool already handles retry with backoff; the sink's internal loop causes up to 16 attempts. The gRPC path incorrectly sets Content-Encoding: zstd which causes double decompression.

---
*PR created automatically by Jules for task [9495995346976104834](https://jules.google.com/task/9495995346976104834) started by @strawgate*

> [!NOTE]
> ### Fix gRPC Content-Encoding header and add retry on 5xx in OTLP sink
> - Fixes compressed gRPC requests to use `grpc-encoding: zstd` instead of `Content-Encoding: zstd`; HTTP requests continue to use `Content-Encoding: zstd`
> - Changes 5xx responses in `OtlpSink.send_payload` to return `SendResult::RetryAfter(5s)` instead of an error, matching the existing 429 behavior
> - Enables HTTP/2 support in the `reqwest` dependency for the `logfwd-output` crate
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4783137.</sup>
> 
>

### DIFF
--- a/crates/logfwd-output/src/otlp_sink.rs
+++ b/crates/logfwd-output/src/otlp_sink.rs
@@ -27,6 +27,8 @@ use super::{BatchMetadata, Compression, str_value};
 const SCOPE_NAME: &[u8] = b"logfwd";
 /// Version emitted in the OTLP `InstrumentationScope.version` field (from Cargo.toml).
 const SCOPE_VERSION: &[u8] = env!("CARGO_PKG_VERSION").as_bytes();
+/// Default retry-after delay in seconds when the server does not send a Retry-After header.
+const DEFAULT_RETRY_AFTER_SECS: u64 = 5;
 
 // ---------------------------------------------------------------------------
 // OtlpSink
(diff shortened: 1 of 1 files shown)


strawgate:
**Review: ⚠️ Two issues to fix.**

1. **Missing 429 handling (High):** The retry loop was removed but `status.is_server_error()` only catches 500-599. HTTP 429 (Too Many Requests) is a 4xx code and will fall through to `Rejected()` instead of `RetryAfter()`. Need to add explicit 429 check before the general client error block:
```rust
if status == StatusCode::TOO_MANY_REQUESTS {
    return Ok(SendResult::RetryAfter(Duration::from_secs(5)));
}
```

2. **Incomplete gRPC fix (Medium):** The PR correctly removes `Content-Encoding: zstd` from gRPC, but doesn't add the correct `grpc-encoding: zstd` header for gRPC compressed payloads. The fix prevents the wrong header but doesn't set the right one.

PR #1251, opened 2026-04-05
fix(checkpoint): harden segment storage — validation, replay caps, and overflow safety

### LINKED ISSUE #1211: fix(checkpoint): segment storage has panic paths, OOM vector, and silent error masking
Four correctness issues in the Phase 1 checkpoint segment format introduced in #1120:

1. **OOM via corrupt files**: `batch_count` from disk is used directly as `Vec::with_capacity()`. A malformed segment with a huge `batch_count` will OOM the process.
2. **Unreadable segments can be created**: `SegmentFile::create()` lets callers supply arbitrary `header.version`/`header.flags` values that `open()` will later reject, producing segments that can never be replayed.
3. **`append()` doesn't enforce `MAX_READ_DATA_SIZE`**: Large payloads can be written but `read_batches()` will always reject them. Write and read limits are inconsistent.
4. **Unsupported LCHK versions silently skipped**: Unknown version bytes map to `NotASegment` and are silently dropped during recovery instead of hard-erroring.
5. **`unwrap`/`expect` panics in production paths**: `SegmentManager::append` unwraps `self.current`; `rotate()` panics if called with no active segment. Both should return `io::Error`.
6. **Permission errors mapped to `NotASegment`**: OS-level read errors cause silent data loss during recovery.

Needs targeted fixes before the checkpoint pipeline is relied on in production.

strawgate:
Fixes #1211

## Summary

This PR hardens segment storage so corrupt on-disk metadata and impossible segment ID states fail closed instead of creating unreadable files, exhausting memory, or leaving partial state behind.

## What this branch fixes

1. Caps `read_batches()` preallocation so a corrupt footer cannot request an enormous `Vec` allocation.
2. Validates segment headers before `SegmentWriter::create()` writes a file, preventing unreadable segments from being created.
3. Enforces the replayability cap in `append()` so we do not write segments that `read_batches()` must later reject.
4. Distinguishes unsupported-but-valid headers and non-`NotFound` I/O failures from `NotASegment`, improving recovery diagnostics.
5. Avoids truncating `bytes_to_hash` on 32-bit platforms.
6. Checks `next_segment_id` overflow before file creation or manager mutation, so overflow leaves no orphaned file or half-open segment.
7. Preserves the underlying seek error details during checksum verification.
8. Replaces the remaining writer-side unchecked size/counter increments with `checked_add` guard rails.

## Additional polish in this pass

- Puts the `read_batches()` docs on the public API instead of on the internal size cap constant.
- Adds focused regression tests for:
  - invalid header version rejection
  - unknown flag rejection
  - append size-cap enforcement
  - no-ZSTD writer path
  - overflow with no side effects through both the public append path and the internal open path

## Verification

- `cargo fmt --all`
- `cargo test -p logfwd-io --lib segment -- --nocapture`
- `cargo clippy -p logfwd-io --lib -- -D warnings`

> [!NOTE]
> ### Harden segment storage with validation, replay caps, and overflow safety
> - Expands segment filename zero-padding from 10 to 20 digits in [`segment.rs`](https://github.com/strawgate/memagent/pull/1251/files#diff-1aab6b5347baf9761af45286454562e02fbc2212a96b37641264b86487e8eff4) to cover the full u64 range and maintain lexicographic sort order
> - `SegmentFile::open` now returns `Corrupt` with detailed diagnostics for most IO/parse failures, and `Incomplete` on `UnexpectedEof`, rather than generic errors
> - `SegmentWriter::create` rejects unsupported versions and unknown flag bits with `InvalidInput`; `append` now respects the `FLAG_ZSTD` header flag for conditional compression and enforces `MAX_READ_DATA_SIZE`
> - `SegmentFile::read_batches` caps pre-allocated batch vector capacity at 65536 to avoid trusting unbounded on-disk counts
> - All u64/usize arithmetic in segment ID generation, size tracking, and checksum computation uses `checked_add` to return errors instead of wrapping on overflow
> - Behavioral Change: segment filenames change from 10-digit to 20-digit padding; existing filenames with 10-digit padding will no longer sort correctly relative to new ones
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8eaee6e.</sup>
> 
>

### DIFF
--- a/crates/logfwd-io/src/segment.rs
+++ b/crates/logfwd-io/src/segment.rs
@@ -408,15 +408,15 @@ impl SegmentFile {
         })
     }
 
-    /// Read all RecordBatches from a validated segment.
-    ///
-    /// Each `append()` wrote an independent IPC stream (schema + batch + EOS),
-    /// so we create a new `StreamReader` for each sub-stream using a shared
-    /// cursor that tracks the read position across streams.
     /// Maximum data_size we'll allocate for reading (1 GiB).
     /// Defense-in-depth against corrupt footer claiming enormous size.
     const MAX_READ_DATA_SIZE: u64 = 1024 * 1024 * 1024;
 
+    /// Read all `RecordBatch` values from a validated segment.
+    ///
+    /// Each `append()` wrote an independent IPC stream (schema + batch + EOS),
+    /// so this replays the segment by opening a fresh `StreamReader` for each
+    /// sub-stream while a shared cursor tracks the byte position.
     pub fn read_batches(&self) -> io::Result<Vec<RecordBatch>> {
         if self.footer.data_size > Self::MAX_READ_DATA_SIZE {
             return Err(io::Error::new(
@@ -588,7 +588,10 @@ impl SegmentWriter {
             .map_err(|e| io::Error::other(e.to_string()))?;
 
         // Enforce the read-time limit so every written segment can be replayed.
-        let new_data_size = self.data_size + self.ipc_buf.len() as u64;
+        let new_data_size = self
+            .data_size
+            .checked_add(self.ipc_buf.len() as u64)
+            .ok_or_else(|| io::Error::other("segment data_size overflow"))?;
         if new_data_size > SegmentFile::MAX_READ_DATA_SIZE {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
@@ -602,8 +605,14 @@ impl SegmentWriter {
         self.writer.write_all(&self.ipc_buf)?;
         self.hasher.update(&self.ipc_buf);
 
-        self.record_count += batch.num_rows() as u64;
-        self.batch_count += 1;
+        self.record_count = self
+            .record_count
+            .checked_add(batch.num_rows() as u64)
+            .ok_or_else(|| io::Error::other("segment record_count overflow"))?;
+        self.batch_count = self
+            .batch_count
+            .checked_add(1)
+            .ok_or_else(|| io::Error::other("segment batch_count overflow"))?;
         self.data_size = new_data_size;
 
         Ok(())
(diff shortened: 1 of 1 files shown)


strawgate:
Two correctness bugs found on review — blocking merge:

**1.  padding insufficient for u64 (low priority)**
`segment_filename` at line 197 uses `{segment_id:010}`, but `segment_id` is `u64` which can have up to 20 digits. Once `segment_id >= 10_000_000_000`, filenames lose lexicographic = numeric sort order (`seg-10000000000.lchk` sorts before `seg-9999999999.lchk`). Fix: `{segment_id:020}`.

**2. ZSTD compression always enabled regardless of `header.flags` (MAJOR)**
`SegmentWriter::append()` (line ~567) unconditionally applies ZSTD compression:
```rust
let opts = IpcWriteOptions::default().try_with_compression(Some(CompressionType::ZSTD))
```
This should check `self.header.flags & FLAG_ZSTD != 0` before enabling compression, otherwise the actual encoding can disagree with the header metadata.

Additionally, this PR has merge conflicts that need to be resolved before it can merge.

strawgate:
@coderabbitai review

Fixed in commit d30c476:
- **CRITICAL: 20-digit padding** — Changed `{segment_id:010}` to `{segment_id:020}` in `segment_filename()`. u64::MAX is 20 digits; 10-digit padding would cause lexicographic sort to diverge from numeric order for IDs ≥ 10_000_000_000, breaking recovery ordering.
- Updated tests to assert 20-digit format and added u64::MAX case.
- Updated comment at the sort call site.

The MAJOR ZSTD concern is a confirmed false positive: every segment is created with `flags: FLAG_ZSTD` unconditionally in `SegmentWriter::create()` (line 753), so there is no code path where `FLAG_ZSTD == 0` for a segment in use.

PR #1265, opened 2026-04-05
fix(bench): separate last/peak RSS atomics; guard iterations=0 panic (#1220)

### LINKED ISSUE #1220: fix(bench): framed_input_profile peak RSS tracking broken — both atomics point to same value
In `crates/logfwd-bench/src/bin/framed_input_profile.rs` (merged in #1125):

```rust
let last_rss_kb = Arc::clone(&peak_rss_kb);
```

Both `last_rss_kb` and `peak_rss_kb` point to the same `AtomicU64`. The sampler reads back the value it just wrote when checking `rss > current_peak`, so `rss > current_peak` is always false. Peak RSS is always equal to the last sample, not the actual peak.

Also: `--iterations 0` causes an out-of-bounds panic at `samples[samples.len() / 2]`.

Fix: use a separate `AtomicU64` for `last_rss_kb`; add CLI validation rejecting `--iterations 0`.

strawgate:
## Summary

Two bugs in `framed_input_profile.rs`:

1. **Peak RSS never tracked correctly:** `Arc::clone(&peak_rss_kb)` made `last_rss_kb` an alias of `peak_rss_kb` — the same `AtomicU64`. The peak check `rss > current_peak` read back the value just written, so it was always false. Fixed by creating a separate `Arc<AtomicU64>` for `last_rss_kb`.

2. **`--iterations 0` panic:** `samples[samples.len() / 2]` panicked with zero iterations. Added a guard via `assert!`.

## Test plan
- [x] `cargo build -p logfwd-bench` passes
- [x] `cargo clippy -p logfwd-bench --bin framed_input_profile` clean
- [ ] CI green

Closes #1220

> [!NOTE]
> ### Fix separate last/peak RSS atomics and guard against iterations=0 panic in bench
> - Fixes `RssSampler.start` in [framed_input_profile.rs](https://github.com/strawgate/memagent/pull/1265/files#diff-cba7aff1a5fddfd5abfa06c66fe752ee02efd643683d1e3bc85ae7ef34b53f0e) so `last_rss_kb` and `peak_rss_kb` are independent atomics; previously `last_rss_kb` mirrored `peak_rss_kb` instead of tracking the most recent sample.
> - Fixes `median_time_ns` to return 0 instead of panicking when the samples vector is empty (i.e. `--iterations=0`).
> - Adds an assertion in `measure_pipeline_scenario` that panics with a clear message (`--iterations must be at least 1`) when iterations is 0.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6431bcf.</sup>
> 
>

### DIFF
--- a/crates/logfwd-bench/src/bin/framed_input_profile.rs
+++ b/crates/logfwd-bench/src/bin/framed_input_profile.rs
@@ -346,6 +346,7 @@ fn measure_pipeline_scenario(scenario: &Scenario, iterations: usize) -> StageSum
         samples.push(run_pipeline_once(scenario));
     }
     samples.sort_by_key(|s| s.total_ns);
+    assert!(!samples.is_empty(), "--iterations must be at least 1");
     let median = &samples[samples.len() / 2];
     StageSummary {
         name: scenario.name,
@@ -467,6 +468,9 @@ fn median_time_ns(iterations: usize, mut f: impl FnMut()) -> u64 {
         samples.push(start.elapsed().as_nanos() as u64);
     }
     samples.sort_unstable();
+    if samples.is_empty() {
+        return 0;
+    }
     samples[samples.len() / 2]
 }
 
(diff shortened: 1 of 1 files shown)


strawgate:
The assert prevents the OOB panic which is the fix for this issue. Moving it into `Cli::parse()` for earlier feedback is a good enhancement — added as a follow-up to keep this PR minimal.

PR #1297, opened 2026-04-05
fix(otlp_sink): return RetryAfter on 5xx server errors

strawgate:
## Problem

When an OTLP endpoint returns a 5xx server error (502 Bad Gateway, 503 Service Unavailable, 504 Gateway Timeout, etc.), the sink previously propagated this as a hard `Err`, bypassing the retry loop. These are transient errors — the server is temporarily overloaded or restarting — and should be retried, not treated as fatal delivery failures.

## Fix

- Add `DEFAULT_RETRY_AFTER_SECS = 5` named constant (replaces the magic `5` in the existing 429 path)
- Add `if status.is_server_error() { return Ok(RetryAfter(DEFAULT_RETRY_AFTER_SECS)) }` before the catch-all `Err`, so the retry loop handles 5xx the same as 429 Too Many Requests
- Update `send_payload_returns_err_on_5xx_no_loop` → `send_payload_returns_retry_after_on_5xx` to match the new (correct) behavior

## Relation to PR #1242

This is the remaining unmerged functional change from PR #1242 (`fix(logfwd-output): OTLP sink retry and gRPC Content-Encoding fixes`). The Content-Encoding fix from that PR already landed in master separately. PR #1242 has accumulated 15+ commits of master drift making a clean rebase impractical; this cherry-picks only the substantive fix.

Closes #1242

> [!NOTE]
> ### Return `RetryAfter` on 5xx responses in `OtlpSink.send_payload`
> - 5xx responses now return `SendResult::RetryAfter` instead of `Err`, using the `Retry-After` header value or a 5-second default.
> - Adds a `parse_retry_after` helper in [otlp_sink.rs](https://github.com/strawgate/memagent/pull/1297/files#diff-56ca7e19db98793b4ef621c0d28a618a4c8af2bd492bc89cc8a21e845f25a43c) that handles both delta-seconds and HTTP-date formats, falling back to `DEFAULT_RETRY_AFTER_SECS` (5s) when the header is absent, unparsable, or in the past.
> - 429 responses also benefit from the new parser, gaining support for HTTP-date values in addition to delta-seconds.
> - Behavioral Change: 5xx responses no longer propagate as errors; callers relying on `Err` for server errors will now receive `RetryAfter` instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7c7df12.</sup>
> 
>

### DIFF
--- a/crates/logfwd-output/src/otlp_sink.rs
+++ b/crates/logfwd-output/src/otlp_sink.rs
@@ -36,7 +36,7 @@ const DEFAULT_RETRY_AFTER_SECS: u64 = 5;
 /// - delta-seconds: `"120"` → `Duration::from_secs(120)`
 /// - HTTP-date: `"Wed, 21 Oct 2015 07:28:00 GMT"` → seconds until that time
 ///
-/// Falls back to `DEFAULT_RETRY_AFTER_SECS` if the value is absent, unparseable,
+/// Falls back to `DEFAULT_RETRY_AFTER_SECS` if the value is absent, unparsable,
 /// or already in the past.
 fn parse_retry_after(header_value: Option<&reqwest::header::HeaderValue>) -> Duration {
     let Some(value) = header_value.and_then(|v| v.to_str().ok()) else {


strawgate:
Addressed CodeRabbit MAJOR: added `parse_retry_after()` helper that tries delta-seconds first, then falls back to `httpdate::parse_http_date()` for the IMF-fixdate format. Both the 429 and 5xx branches now call the helper. `httpdate` was already a transitive dep via reqwest — added it as explicit direct dep.

Issue #1363, opened 2026-04-06
RESEARCH: processor chain architecture — I/O/compute separation, pipeline compiler, multi-pipeline topology


strawgate:
## Summary

Deep architecture research for the next-generation processor chain in logfwd. This replaces the monolithic `pipeline.rs` select loop with a multi-pipeline topology where inputs, processors, and outputs are independent mini-pipelines connected by bounded channels.

**Core principle: complete separation of I/O and compute.**

## Architecture

```
Input Pipeline (per input)          Processor Pipeline           Output Pipeline (per output)
┌──────┐  ch  ┌──────┐     ch     ┌──────────────────┐   ch    ┌──────┐  ch  ┌──────┐
│ I/O  │─────▶│ CPU  │───────────▶│ Sync processors  │────────▶│ CPU  │─────▶│ I/O  │
│Worker│      │Worker│            │ (user-configured) │         │Worker│      │Worker│
│      │      │      │            │                   │         │      │      │      │
│read()│      │scan()│            │ [checkpoint]      │         │sql() │      │POST  │
│accum │      │sql() │            │ [tail_sampling]   │         │enc() │      │retry │
└──────┘      └──────┘            └──────────────────┘         └──────┘      └──────┘
 OS thread    spawn_blocking       tokio task                   spawn_blocking tokio task
```

### 1. I/O and compute on separate workers
- **Prototype**: Split mode achieves 2× throughput of fused by overlapping I/O and CPU
- **Correctness**: I/O code can never accidentally block compute, and vice versa
- **Testability**: Each worker is independently testable with mocks
- The compiler can **fuse** workers when separation isn't needed (simple file inputs with no SQL)

### 2. Sync Processor trait
```rust
pub trait Processor: Send {
    fn process(&mut self, batch: RecordBatch, meta: &BatchMetadata)
        -> Result<SmallVec<[RecordBatch; 1]>, ProcessorError>;
    fn flush(&mut self) -> SmallVec<[RecordBatch; 1]>;
    fn name(&self) -> &str;
    fn is_stateful(&self) -> bool;
}
```
- >90% of processors are CPU-bound. Async adds Pin<Box<dyn Future>> allocation for no benefit.
- Checkpoint (the one I/O processor) uses internal channel + background task — proven pattern from input threads.
- SmallVec<[RecordBatch; 1]> avoids heap allocation for 99% case (1:1 mapping).

### 3. Pipeline compiler with fusion
[...]

PR #1402, opened 2026-04-06
feat(pipeline): evolve Processor trait to chain with error handling, flush, and SmallVec output

strawgate:
## Summary

Evolves the minimal `Processor` trait (from PR #1322) into a full chain execution framework — the foundation for all future processors (tail sampling, checkpoint, add_field, dedup, etc.).

### Processor trait (`crates/logfwd/src/processor/mod.rs`)
- **`ProcessorError`** enum: `Transient` (retry, don't advance checkpoint), `Permanent` (skip, ack but don't forward), `Fatal` (shutdown without advancing checkpoint). Fixes current `reject()` which silently loses data by advancing checkpoint.
- **`process(batch, meta)`** → `Result<SmallVec<[RecordBatch; 1]>, ProcessorError>` — multi-batch output, proper error handling
- **`flush()`** — graceful shutdown, emit ALL buffered state
- **`name()`** / **`is_stateful()`** — metrics and heartbeat behavior
- **`run_chain()`** — execute batch through ordered processor chain with SmallVec (zero heap alloc for 1:1 case)
- **`cascading_flush()`** — shutdown flush with re-flush: step i's flush feeds step i+1 via process(), then i+1 re-flushes to handle cascade data

### Pipeline integration (`crates/logfwd/src/pipeline.rs`)
- `Option<Box<dyn Processor>>` → `Vec<Box<dyn Processor>>` (chain support)
- `run_chain()` replaces single post-transform call, with error classification per `ProcessorError` variant
- Multi-batch output consolidated via `concat_batches`
- Heartbeat in `flush_interval` arm for stateful processors (sends empty batch through chain)
- Cascading flush in shutdown path
- `with_processor()` pushes to chain; new `with_processors()` for bulk add

### Dependencies
- Added `smallvec` (workspace) and `thiserror` (workspace) to `logfwd`

## Design decisions

- **Sync trait**: >90% of processors are CPU-bound. The one that needs async I/O (checkpoint) will use internal channels + background tasks — same proven pattern as input threads.
- **No tick()**: Stateful processors check internal timers on every `process()` call. Pipeline sends empty batch heartbeat on flush_interval when idle.
- **SmallVec<[RecordBatch; 1]>**: 56 bytes inline, zero heap alloc for the 99% case (one batch in, one batch out).

## Tests

- 10 new unit tests: chain execution, filter, heartbeat passthrough, cascading flush (single + chained stateful), error propagation (3 types)
- Updated `test_pipeline_with_processor` for new 4-method trait signature
- All 11 tests pass

## Known TODOs

- Heartbeat and cascading flush output not yet submitted to output pool (needs ticketless submission — no `BatchTicket` available for timer-triggered data). Logged with `tracing::debug`/`tracing::info`. Will be resolved when first stateful processor (TailSampling) is implemented.
- Fatal error should trigger pipeline shutdown (currently just returns early)

## Test plan

```bash
cargo test -p logfwd --lib -- processor   # 10 new tests
cargo test -p logfwd --lib -- test_pipeline_with_processor  # updated integration test
cargo clippy -p logfwd --lib -- -D warnings  # zero warnings
```

Relates to #1363

> [!NOTE]
> ### Evolve `Processor` trait to support chaining, error classification, flush, and `SmallVec` output
> - Redesigns the `Processor` trait in [mod.rs](https://github.com/strawgate/memagent/pull/1402/files#diff-30ce1c709d62c0e50dc143f58533ce978f7bb33c1fbb800074a32fd635f5fbf5): `process` now returns `Result<SmallVec<[RecordBatch;1]>, ProcessorError>` and the trait gains `flush()`, `name()`, and `is_stateful()` methods.
> - Introduces `ProcessorError` with `Transient`, `Permanent`, and `Fatal` variants (via `thiserror`) so callers can distinguish recoverable from fatal failures.
> - Adds `run_chain` to execute a batch through an ordered sequence of processors, and `cascading_flush` to drain buffered state across the chain in dependency order during shutdown.
> - Replaces the single optional processor field in `Pipeline` with a `Vec<Box<dyn Processor>>` chain; `process_batch` now runs the full chain and handles each error variant (reject/ack/shutdown) accordingly.
> - Adds heartbeat and cascading-flush paths to the `Pipeline::run_async` loop — emitted rows are currently only logged and not submitted to the output pool.
> - Risk: stateful processors registered via `with_processor` or `with_processors` now panic at registration time; previously this was a no-op.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ad677d2.</sup>
> 
>

### DIFF
--- a/crates/logfwd/src/pipeline.rs
+++ b/crates/logfwd/src/pipeline.rs
@@ -421,7 +421,18 @@ impl Pipeline {
     }
 
     /// Add a post-transform processor to the chain.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `processor.is_stateful()` returns `true`. Stateful processors
+    /// require deferred-ACK checkpointing support that is not yet implemented
+    /// (tracked in #1404). Register only stateless processors until then.
     pub fn with_processor(mut self, processor: Box<dyn Processor>) -> Self {
+        assert!(
+            !processor.is_stateful(),
+            "stateful processors are not yet supported: checkpointing path is incomplete \
+             (see #1404). Register only stateless processors."
+        );
         self.processors.push(processor);
         self
     }
@@ -429,7 +440,19 @@ impl Pipeline {
     /// Append a chain of post-transform processors to any already registered via
     /// `with_processor`.  Calling `with_processor(a).with_processors(vec![b, c])`
     /// produces the three-stage chain `[a, b, c]`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if any processor in `processors` returns `is_stateful() == true`.
+    /// See [`with_processor`](Self::with_processor) for details.
     pub fn with_processors(mut self, processors: Vec<Box<dyn Processor>>) -> Self {
+        for p in &processors {
+            assert!(
+                !p.is_stateful(),
+                "stateful processors are not yet supported: checkpointing path is incomplete \
+                 (see #1404). Register only stateless processors."
+            );
+        }
         self.processors.extend(processors);
         self
     }
--- a/dev-docs/research/per-input-sql-analysis.md
+++ b/dev-docs/research/per-input-sql-analysis.md
@@ -365,13 +365,20 @@ If an input has `sql:`, it runs **before** the pipeline-level `transform:`. If a
 
 ### Current state: there is no step compiler
 
-Today's pipeline is a fixed linear chain:
+Today's pipeline is a fixed linear chain (as of the original research; a
+post-transform `Processor` chain was added in PR #1402):
 
 ```text
-Inputs → Scanner → SqlTransform → Outputs
+Inputs → Scanner → SqlTransform → [Processors...] → Outputs
 ```
 
-There is no step chain, no step compiler, no step types. The `SqlTransform` **is** the entire transform layer. All filtering, projection, renaming, enrichment joins, and computed columns are expressed as SQL.
+The `SqlTransform` **is** the primary transform layer for user-visible SQL.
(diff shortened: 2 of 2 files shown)


strawgate:
Fix pushed (commit e59b1a8) addressing the CodeRabbit CRITICAL flag:

**"Don't ack checkpoints when total_rows == 0 means 'buffered internally'"**

Added a \`debug_assert\` that fires in test builds if stateful processors are ever present when the early-ACK path is taken. Also expanded the comment to document the invariant explicitly:

- ACKing on \`total_rows == 0\` is safe today because no stateful processors exist yet.
- When stateful processors land, the two cases (SQL-filtered vs buffered) must be distinguished — the \`debug_assert\` will catch any premature merge.
- \`TODO(#1404)\` references the full fix (ticketless submission + deferred ACK).

Previous fix summary:
1. ✅ \`with_processors\` uses \`extend\` not assignment (f989e2e)
2. ✅ All \`TODO(#0)\` → \`TODO(#1404)\` (f989e2e)
3. ✅ Checkpoint unsafe-ACK guarded by \`debug_assert\` (e59b1a8)

@coderabbitai review

strawgate:
The concat_batches panic path is already handled in the latest commit (ad677d2e) with a proper error/reject path. The remaining open CR items (ticketless submission for cascading_flush, Fatal error → pipeline shutdown, empty-batch checkpoint safety) are tracked in #1404 and have TODO comments with debug_assert guards in place. CI is green. The stale CHANGES_REQUESTED from CodeRabbit is from before the latest fix commit — requesting CR re-review.

PR #1558, opened 2026-04-07
fix(dashboard): stabilize trace explorer, rate-limit backpressure logs

strawgate:
## Summary

Fixes the trace explorer resetting on every poll, reduces noisy backpressure warnings, and removes dead code.

### Trace explorer stability
- `setTraces()` in app.tsx now merges by `trace_id` instead of replacing the array wholesale. This preserves referential stability so the TraceExplorer component does not lose scroll position, expanded state, or re-render unnecessarily on each 2-second poll.
- Tightened batch latency filter to explicitly check `total_ns > 0` for completed traces.

### Backpressure log rate-limiting
- All 3 `input.backpressure` WARN sites in `input_pipeline.rs` are now rate-limited to once per 5 seconds per input.
- `metrics.inc_backpressure_stall()` still fires on every occurrence — only the log is throttled.

### Dead code removal
- Deleted `dashboard/src/lib/stats.ts` — duplicated `computeStats` and `buildLanes` from TraceExplorer, never imported by any production code.
- Exported `computeStats` from TraceExplorer so existing tests can import it.
- Updated test imports to point to TraceExplorer instead of deleted file.

## Test plan
- `cd dashboard && npx vitest run` — 178 tests pass
- `cargo clippy -p logfwd` — zero warnings
- Live verified: started forwarder with generator → OTLP sink, confirmed trace explorer no longer resets, backpressure log output reduced from hundreds to ~3 per minute

> [!NOTE]
> ### Fix trace explorer re-renders and rate-limit backpressure logs
> - Traces state updates in [App.tsx](https://github.com/strawgate/memagent/pull/1558/files#diff-89253c48a25d00a0657e8bf6c6803d4d8d86374de5dbcca18905f910e2cd8255) and [dashboard.html](https://github.com/strawgate/memagent/pull/1558/files#diff-3ac0522b03b81482ecfd1f916c43a39b98d658c5cb2d271f65e6b6eef153af12) now reuse existing objects when render-affecting fields (`total_ns`, `status`) are unchanged, avoiding unnecessary React re-renders on polling.
> - Batch latency calculation now filters out in-progress traces and those with `total_ns <= 0` before averaging.
> - Backpressure warnings in [input_pipeline.rs](https://github.com/strawgate/memagent/pull/1558/files#diff-bf7cf8ae1b6fe98bd6b6fd6db4fa133b060008cef848a465f84c294604ad6c96) are rate-limited to at most once every 5 seconds per input; metrics and blocking behavior are unchanged.
> - `computeStats` is moved from the deleted `lib/stats.ts` into `TraceExplorer.tsx` and re-exported; test imports are updated accordingly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3a3f08d.</sup>
> 
>

### DIFF
--- a/dashboard/src/app.tsx
+++ b/dashboard/src/app.tsx
@@ -294,11 +294,12 @@ export function App() {
         // Build next array keyed by trace_id, reusing previous objects when all
         // render-affecting fields are unchanged to preserve referential stability.
         const prevById = new Map(prev.map((t) => [t.trace_id, t]));
-        let anyChanged = false;
         const next = incoming.map((t) => {
           const p = prevById.get(t.trace_id);
           if (
             p &&
+            p.pipeline === t.pipeline &&
+            p.start_unix_ns === t.start_unix_ns &&
             p.total_ns === t.total_ns &&
             p.status === t.status &&
             p.in_progress === t.in_progress &&
@@ -326,10 +327,10 @@ export function App() {
           ) {
             return p;
           }
-          anyChanged = true;
           return t;
         });
-        if (!anyChanged && next.length === prev.length) return prev;
+        // Always return `next`: it preserves referential stability for unchanged
+        // objects (via prevById lookup) while also reflecting server-side reordering.
         return next;
       });
     }


strawgate:
**Re: CodeRabbit round 4 CHANGES_REQUESTED**

The outstanding item from the 4th-round review — adding `pipeline` and `start_unix_ns` to the trace reuse-check — was addressed in commit `3a3f08d8` (pushed after the CR review was posted). Both fields are now included in the equality guard at lines 301–302 in `app.tsx`.

The ordering-stale issue (comment 3048268967, already auto-marked ✅ Resolved by CR) was also fixed in the same commit by removing the early-return optimization entirely; the function now always returns `next`, which already preserves object identity for unchanged traces via the `prevById` lookup.

@coderabbitai: all actionable items from the 4th round are resolved. Please re-review and dismiss the CHANGES_REQUESTED if no new issues are found.

Issue #1592, opened 2026-04-08
bug: grok() and geo_lookup() struct output serialized as empty string "" in JSON


strawgate:
## Summary

The `grok()` UDF (and any UDF returning a non-conflict `Struct`) is silently serialized as an empty string `""` in all JSON output modes. Users following the documented example pattern see their parsed data silently discarded.

## Reproduction

```bash
# Generate test data
./logfwd --generate-json 20 /tmp/logs.json

# Config: config.yaml
cat > /tmp/config.yaml <<'EOF'
input:
  type: file
  path: /tmp/logs.json
  format: json

transform: |
  SELECT
    level,
    grok(message, '%{NOTSPACE:a} %{NOTSPACE:b} %{NOTSPACE:method} %{URIPATH:path}/%{NUMBER:id}') AS parsed
  FROM logs
  WHERE level = 'ERROR'
  LIMIT 3

output:
  type: stdout
  format: json
EOF

./logfwd --config /tmp/config.yaml
```

## Actual output

```
{"level":"ERROR","parsed":""}
{"level":"ERROR","parsed":""}
{"level":"ERROR","parsed":""}
```

## Expected output

```json
{"level":"ERROR","parsed":{"method":"GET","path":"/health","id":"10021"}}
{"level":"ERROR","parsed":{"method":"GET","path":"/api/v2/products","id":"10049"}}
```

## Root cause

`write_json_value` in `crates/logfwd-output/src/lib.rs` does not have a case for `DataType::Struct`. It falls through to the wildcard `_` branch which calls `str_value(arr, row)`. `str_value` returns `""` for Struct types, so the struct content is silently lost.

The affected code path:
- `write_json_value` (lib.rs ~line 436): `_ => write_json_string(out, str_value(arr, row))?`
- `str_value` (lib.rs ~line 336): `_ => ""`

This affects any SQL query that produces a non-conflict Struct column in the output batch: `grok()`, `geo_lookup()`, or any user SQL that produces a struct-typed column.

## Workaround

Use `get_field()` to extract individual fields from the struct:
```sql
SELECT level,
  get_field(grok(message, '...'), 'method') AS method,
  get_field(grok(message, '...'), 'id') AS id
FROM logs
```

## Fix

Add a `DataType::Struct(child_fields)` match arm to `write_json_value` that recursively serializes the struct as a JSON object.

This bug affects all JSON-based outputs: stdout (json/console format), file (json format), Elasticsearch, OTLP JSON, HTTP JSON lines.

PR #1621, opened 2026-04-08
fix(output): render Struct columns as JSON in console format (#1620)

### LINKED ISSUE #1620: bug: console format silently drops struct column values (grok/geo_lookup show empty)
## Summary

The \`console\` output format silently discards data from any Struct-typed column (e.g., the output of \`grok()\` or \`geo_lookup()\`). Fields appear as \`key=\` with an empty value instead of the actual struct content.

Note: the related issue #1592 mentions console format, and PR #1595 fixes the JSON serialization path in \`write_json_value()\`, but the **console format uses a completely separate code path** (\`write_console()\` in \`stdout.rs\`) that has an identical bug and is not fixed by #1595.

## Root Cause

\`write_console()\` at \`crates/logfwd-output/src/stdout.rs:237-252\` dispatches on Arrow DataType:

\`\`\`rust
match arr.data_type() {
    DataType::Int64 => { /* ... */ }
    DataType::Float64 => { /* ... */ }
    _ => {
        self.buf.extend_from_slice(str_value(arr, row).as_bytes());
    }
}
\`\`\`

The \`_ =>\` arm calls \`str_value()\` which returns \`""\` for any non-string type including \`DataType::Struct\`. There is no \`DataType::Struct\` arm.

## Reproduction

\`\`\`bash
./logfwd --generate-json 20 /tmp/logs.json

cat > /tmp/config.yaml <<'CONF'
input:
  type: file
  path: /tmp/logs.json
  format: json

transform: |
  SELECT
    level,
    message,
    grok(message, '%{NOTSPACE:a} %{NOTSPACE:b} %{NOTSPACE:method} %{URIPATH:path}/%{NUMBER:id}') AS parsed
  FROM logs
  WHERE level = 'ERROR'
  LIMIT 3

output:
  type: stdout
  format: console
CONF

./logfwd --config /tmp/config.yaml
\`\`\`

## Expected output

\`\`\`
2024-01-15T10:30:00Z  ERROR  request handled GET /api/v2/products/10021  parsed={"a":"ERROR","b":"request","method":"GET","path":"/api/v2/products","id":"10021"}
\`\`\`

## Actual output

\`\`\`
2024-01-15T10:30:00Z  ERROR  request handled GET /api/v2/products/10021  parsed=
\`\`\`

## Impact

- Any \`console\` format output using \`grok()\` or \`geo_lookup()\` silently loses all struct data
- No error or warning is emitted — data loss is invisible to the user
- The same pattern works with \`format: json\` after #1595 is merged, but \`format: console\` remains broken

## Fix

Add a \`DataType::Struct\` arm in \`write_console()\` that emits the struct as inline JSON (matching how the json format represents it post-#1595).

strawgate:
## Summary

- Adds `DataType::Struct` arm to `write_console()` in `stdout.rs`, serialising struct columns (e.g. `grok()`, `geo_lookup()`) as inline JSON (`parsed={"method":"GET","status":"200"}`)
- Makes `write_json_value()` `pub(crate)` so `write_console()` can call it for recursive struct serialisation
- Adds `DataType::Struct` handler to `write_json_value()` itself (same change as pending PR #1595; included here so the fix works independently of merge order, and will be a clean no-op on merge)

## Root cause

`write_console()` dispatched `Int64` and `Float64` explicitly, then fell through to `str_value()` for all other types. `str_value()` returns `""` for any non-string Arrow type (including `DataType::Struct`). There was no `DataType::Struct` arm.

## Why this is separate from #1592 / #1595

Issue #1592 covers the JSON serialisation path (`write_json_value()` via `write_row_json()`). PR #1595 fixes that path. The **console format** uses a completely separate code path (`write_console()` in `stdout.rs`) that PR #1595 does not touch. This PR fixes that second path.

## Test plan

Two new regression tests added to `crates/logfwd-output/src/lib.rs`:

- [ ] `test_console_format_struct_column_rendered_as_json` — asserts non-null struct fields appear as JSON in console output (would have failed before this fix: showed `parsed=""`)
- [ ] `test_console_format_null_struct_cell` — asserts null struct rows are correctly skipped

- [ ] `cargo test -p logfwd-output` — all 171 tests pass
- [ ] `cargo clippy -p logfwd-output -- -D warnings` — clean
- [ ] `cargo fmt -p logfwd-output` — clean

Closes #1620

> [!NOTE]
> ### Render Struct columns as inline JSON in console output format
> Previously, Struct-typed columns appeared as empty strings in stdout console output. `StdoutSink.write_batch_to` now matches on `DataType::Struct(_)` and serializes the cell using `write_json_value` before falling back to `str_value` for other types. Two regression tests cover both non-null and null Struct cells.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6037793.</sup>
> 
>

### DIFF
--- a/crates/logfwd-io/src/platform_sensor_beta.rs
+++ b/crates/logfwd-io/src/platform_sensor_beta.rs
@@ -94,10 +94,10 @@ impl PlatformSensorBetaInput {
 
 impl InputSource for PlatformSensorBetaInput {
     fn poll(&mut self) -> io::Result<Vec<InputEvent>> {
-        let should_emit = if !self.started {
-            true
-        } else {
+        let should_emit = if self.started {
             self.cfg.emit_heartbeat && self.last_emit.elapsed() >= self.cfg.poll_interval
+        } else {
+            true
         };
 
         if !should_emit {


strawgate:
## Review: Blocking issues

**CI is broken** — the config refactor removed types/fields that `logfwd-runtime` depends on (`GeneratorAttributeValueConfig`, `GeneratorComplexityConfig`, `GeneratorProfileConfig`, `HttpMethodConfig`, `InputConfig.sql`, `InputConfig.generator`, `InputType::Http`, etc.), causing 10+ compile errors.

**Scope concern** — the PR description says it fixes console Struct rendering (#1620), but ~2100 of ~2400 diff lines are an unrelated config crate refactor (inlining 6 modules into one `lib.rs`). The actual console fix is ~50 lines across `stdout.rs` and `row_json.rs` and looks correct.

**Recommendation:**
1. Strip the config refactor from this PR, keeping only the Struct rendering fix
2. Or split into two PRs: one for the console fix (small, reviewable) and one for the config refactor (needs its own review and must preserve the types that `logfwd-runtime` uses)

The Struct rendering code itself (recursive `write_json_value` for nested structs, safe downcast with `let Some(...) else`) is well-done. Note that PR #1617 also adds Struct/Binary rendering to `row_json.rs` — these will need to be reconciled.

PR #1637, opened 2026-04-08
fix(config): validate workers >= 1 and batch_target_bytes > 0 at --validate time

### LINKED ISSUE #1636: bug: config validator accepts workers: 0 and batch_target_bytes: 0 that fail at runtime
## Summary

Two pipeline configuration fields are validated at runtime (in `pipeline.rs`) but not at `--validate` time (in `logfwd-config`). This means `logfwd --validate` reports success for configurations that will fail immediately when the pipeline starts.

## Affected Fields

| Field | Runtime check | `--validate` check |
|-------|--------------|---------------------|
| `pipeline.workers` | `pipeline.rs:142` — "workers must be >= 1" | ❌ missing |
| `pipeline.batch_target_bytes` | `pipeline.rs:145` — "batch_target_bytes must be > 0" | ❌ missing |

Compare with `batch_timeout_ms`, which IS caught at validate time (`lib.rs:626`).

## Reproduction

```yaml
pipelines:
  main:
    workers: 0        # passes --validate, fails at runtime
    inputs: [...]
    outputs: [...]
```

`logfwd --validate --config above.yaml` prints `Config is valid` but `logfwd --config above.yaml` fails immediately with: `workers must be >= 1`.

## Fix

Add the same guard to `validate()` in `crates/logfwd-config/src/lib.rs` that already exists in `pipeline.rs`:

```rust
// In the per-pipeline loop in validate():
if pipe.workers == Some(0) {
    return Err(ConfigError::Validation(format!(
        "pipeline '{name}': workers must be >= 1"
    )));
}
if pipe.batch_target_bytes == Some(0) {
    return Err(ConfigError::Validation(format!(
        "pipeline '{name}': batch_target_bytes must be > 0"
    )));
}
```

strawgate:
Closes #1636. Mirrors the runtime checks from pipeline.rs into validate() so --validate catches these invalid values before startup.

> [!NOTE]
> ### Validate `workers >= 1` and `batch_target_bytes > 0` in pipeline config at validate time
> Adds two validation checks in `Config::from_raw` in [lib.rs](https://github.com/strawgate/memagent/pull/1637/files#diff-8ce51c6580ec7211876aab0f3aa4c8dece56cbaf454430aa08a8b0992e8f3150): pipelines with `workers: 0` or `batch_target_bytes: 0` now return an explicit error at `--validate` time rather than silently accepting invalid values. Two corresponding tests cover each rejected case.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 58d8111.</sup>
> 
>

### DIFF
--- a/crates/logfwd-config/src/lib.rs
+++ b/crates/logfwd-config/src/lib.rs
@@ -628,6 +628,16 @@ impl Config {
                     "pipeline '{name}': batch_timeout_ms must be greater than 0"
                 )));
             }
+            if pipe.workers == Some(0) {
+                return Err(ConfigError::Validation(format!(
+                    "pipeline '{name}': workers must be >= 1"
+                )));
+            }
+            if pipe.batch_target_bytes == Some(0) {
+                return Err(ConfigError::Validation(format!(
+                    "pipeline '{name}': batch_target_bytes must be > 0"
+                )));
+            }
             if let Some(sql) = &pipe.transform {
                 if sql.trim().is_empty() {
                     return Err(ConfigError::Validation(format!(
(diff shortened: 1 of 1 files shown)


strawgate:
Closing as superseded: the workers >= 1 and batch_target_bytes > 0 validations were incorporated into `crates/logfwd-config/src/validate.rs` during the config module split (#1600). Main already has this fix.

PR #1825, opened 2026-04-12
refactor: simplify I/O ingress paths

strawgate:
## Summary

This PR simplifies and hardens the I/O ingress layer:

- Simplifies OTLP receiver decoding/projection flow and removes the retired projection module.
- Updates HTTP, Arrow IPC, OTAP, generator, UDP, TCP, and tail ingress paths to use the current format, backpressure, and checkpoint contracts.
- Restores same-inode fingerprint rotation detection so copytruncate/rewrite-style rotations are not missed after the fingerprint window.
- Refreshes ingress integration tests for checkpoint state, transport behavior, backpressure, and tail rotation semantics.
- Keeps logfwd-io feature declarations aligned with the remaining code.

## Validation

- `cargo fmt --check`
- `cargo test -p logfwd-io tail::discovery -- --nocapture`
- `cargo test -p logfwd-io -- --nocapture`
- `cargo clippy -p logfwd-io --tests -- -D warnings`

> [!NOTE]
> ### Simplify I/O ingress paths by relaxing content-type validation and enabling integration tests by default
> - HTTP endpoints (`otap_receiver`, `arrow_ipc_receiver`) no longer reject requests based on Content-Type; content-type is now only used to detect encoding hints like `+zstd`
> - `parse_content_encoding` across multiple receivers now trims whitespace, handles comma-separated tokens, and returns HTTP 400 for blank or unsupported values
> - File tail rotation now triggers when the previously held file handle is deleted (even if device/inode/fingerprint are unchanged), and evicted offsets beyond file size are preserved to enable truncation detection instead of resetting to offset 0
> - `AdaptivePollController.observe_signal` counts down an existing burst window instead of re-arming to max on consecutive saturated polls
> - OTLP JSON decoding replaces byte-level hex decoding with `normalize_otlp_hex_id`, which validates and lowercases hex trace/span IDs using string operations
> - All previously `#[ignore]`-d network integration tests in `transport_e2e.rs` now run by default
> - Behavioral Change: Arrow IPC endpoint returns 400 (not 415) for invalid or unsupported Content-Encoding values
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9d48b60.</sup>
> 
>


strawgate (commented):
## AI-Assisted Review (fastmcp-pr-review v2)

### Medium: Missing Retry-After header in 429/503 responses (http_input.rs)

Backpressure and unavailability responses no longer include the `Retry-After` header, which clients use to determine when to retry.

### Low: Removed trim() in parse_content_encoding (http_input.rs)

`trim()` was removed from header parsing. Per RFC 7230, header values can have optional whitespace — `Content-Encoding: gzip ` (trailing space) will now fail to match.

### Medium: proptest failure persistence disabled (checkpoint_state_machine.rs)

Disabling proptest persistence means previously-found failure cases won't be replayed, reducing regression coverage.

---

*Reviewed by [fastmcp-pr-review](https://github.com/strawgate/fastmcp-pr-review) v2*

Issue #1800, opened 2026-04-11
bug: crash+bounce path can exhaust retries and lose post-recovery delivery


strawgate:
## Summary
`tcp_server_crash_triggers_reconnect` currently documents that after server crash+bounce, post-restart delivery may not recover because retry exhaustion occurs during downtime.

## Evidence
- Test: `crates/logfwd/tests/turmoil_sim/network_sim.rs`
- Function: `tcp_server_crash_triggers_reconnect`
- Existing in-test finding states:
  - worker can exhaust retries during crash window
  - after bounce, no new delivery may occur
  - assertion is weakened to `total_received >= pre_crash`

So the test currently accepts "no post-bounce progress" as valid.

## Why this is a bug
Transient downstream outages should not permanently suppress forward progress for the remainder of a run once the destination recovers. Accepting only pre-crash durability masks a recovery gap and weakens the intended reconnect contract.

## Proposed direction
- Strengthen runtime retry/recovery semantics after transport recovery:
  - avoid terminally abandoning active stream after short outage windows
  - ensure worker/pool state can re-enter healthy send path after reconnect
- Strengthen the turmoil contract test:
  - require `total_received > pre_crash` after bounce under bounded outage duration
  - keep deterministic replay seed and trace checks
- Add bounded liveness assertion for "recover-after-bounce" in TLA/turmoil alignment docs.

## Priority rationale
P1 because this is a concrete resilience regression in crash/recovery conditions and can cause silent under-delivery until restart.

PR #1851, opened 2026-04-13
fix: tolerate held batches without halting pipeline ingestion

### LINKED ISSUE #1799: bug: worker panic path can complete run with zero post-panic delivery
## Summary
`worker_panic_does_not_block_drain` currently treats `delivered == 0` as acceptable after a sink panic, as long as the pipeline does not deadlock.

## Evidence
- Test: `crates/logfwd/tests/turmoil_sim/bug_hunt.rs`
- Function: `worker_panic_does_not_block_drain`
- Comment/assert behavior currently allows this outcome:
  - "0 is acceptable if worker pool doesn't respawn after panic"

That means we accept a terminal state where the run finishes but all remaining batches are effectively lost until restart/replay.

## Why this is a bug
A sink worker panic should not degrade the system into "drain completed, zero forward progress" for the rest of the run. We should either:
1. deterministically respawn capacity and continue progressing, or
2. fail the run in a way that preserves explicit replay semantics and observable failure state.

The current acceptance criterion hides a real reliability gap.

## Proposed direction
- Promote the current test from "documented limitation" to an explicit failure contract:
  - require at least one successful post-panic recovery send OR explicit pipeline failure signal.
- Add worker-pool panic recovery invariant tests:
  - worker slot replacement occurs within bounded time
  - queued work eventually terminalizes (ack/reject/hold)
- Add diagnostics metric/span for panic recovery path so this cannot be silent.

## Priority rationale
P1 because this is a user-visible reliability failure mode under fault conditions and can silently mask total post-panic delivery stall.

strawgate:
## Summary

Previously, any single held batch (from retry exhaustion, worker panic, or transient output failure) immediately stopped the entire pipeline — cancelling ingestion and preventing all forward progress. This made transient output failures fatal.

Now the pipeline continues processing while held batches accumulate, only stopping when a safety-valve threshold (10 batches) is reached. Held batches still block checkpoint advancement, preserving at-least-once delivery semantics — on restart, data replays from the last checkpoint.

**Why this works:** when the pipeline continues submitting, the worker pool naturally prunes dead workers and spawns fresh ones with new connections. This allows recovery from both worker panics (#1799) and output bounce scenarios (#1800).

**Tradeoff:** batches delivered after a held batch produce duplicates on restart (checkpoint is blocked at the held gap). With threshold=10, this bounds the duplicate window to ~10 batches before the pipeline concludes the output is persistently broken.

## What changed

**Pipeline core** (`logfwd-runtime/src/pipeline/mod.rs`, `build.rs`):
- `apply_pool_ack()` now returns `true` (stop) only when held batch count reaches `max_held_batches` (default 10), rather than on the first hold
- Added `max_held_batches` field to Pipeline struct with safety-valve semantics

**Turmoil simulation tests** (`bug_hunt.rs`, `network_sim.rs`):
- `worker_panic_does_not_block_drain`: uses 200 lines with 64-byte batch target to guarantee multiple batches; now asserts `delivered > 0` (worker pool respawns after panic)
- `panic_first_batch` and `panic_after_initial_success`: updated from terminal-hold assertions to hold-resilience assertions while preserving checkpoint safety invariants
- `held_pool_ack_requests_terminal_shutdown`: sets `max_held_batches=1` to test safety-valve at lowest threshold

## Safety invariants preserved

- Held batches block checkpoint advancement (at-least-once delivery via replay on restart)
- Checkpoint monotonicity and durable-not-ahead-of-updates
- Safety valve at 10 held batches prevents unbounded memory growth

## Future work (separate PRs)

- **Worker-level fresh-sink retry:** create new sink from factory after panic/retry-exhaustion and retry the batch before acking Hold. This would prevent holds entirely for transient failures but requires TLA+ model updates and proptest coverage.
- Make `max_held_batches` configurable via pipeline YAML config.

## Test plan

- `just test` — 1404 tests pass
- `just clippy` — zero warnings
- All 62 turmoil simulation tests pass (including strengthened panic/crash tests)

Closes #1799
Relates to #1800

> [!NOTE]
> ### Allow pipeline ingestion to continue until held batches reach a configurable threshold
> - Previously, `Pipeline.apply_pool_ack` stopped ingestion on the first held (undeliverable) batch. It now stops only when `held_tickets.len() >= max_held_batches`, defaulting to 10 via `DEFAULT_MAX_HELD_BATCHES`.
> - An error is logged when the safety valve trips. Warning logs for `ack_all_tickets` are updated to reflect that the pipeline continues processing while checkpoints are blocked.
> - Integration tests in [bug_hunt.rs](https://github.com/strawgate/memagent/pull/1851/files#diff-f70ec9a220356a8d2b23628fe22853c0140bc9cc729445ee2ae2be7ab32e90bc) are updated to assert continued delivery after a worker panic, matching the new behavior.
> - Behavioral Change: pipelines that previously halted ingestion on the first held batch will now continue until 10 batches are held.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 606f388.</sup>
> 
>

### DIFF
--- a/crates/logfwd-runtime/src/pipeline/build.rs
+++ b/crates/logfwd-runtime/src/pipeline/build.rs
@@ -41,7 +41,11 @@ pub(crate) const DEFAULT_POOL_DRAIN_TIMEOUT: Duration = Duration::from_secs(60);
 /// accumulate — this allows recovery when the failure is transient (e.g.,
 /// downstream bounce). Only when the held count exceeds this limit does the
 /// pipeline stop accepting new input to bound memory growth.
-pub(crate) const DEFAULT_MAX_HELD_BATCHES: usize = 100;
+///
+/// At 10 batches with typical retry backoff (~2-3 seconds per batch), persistent
+/// failures are detected within ~30 seconds while transient blips (1-3 holds)
+/// recover naturally.
+pub(crate) const DEFAULT_MAX_HELD_BATCHES: usize = 10;
 
 impl Pipeline {
     /// Construct a pipeline from parsed YAML config.
--- a/crates/logfwd-runtime/src/pipeline/build.rs
+++ b/crates/logfwd-runtime/src/pipeline/build.rs
@@ -34,6 +34,15 @@ pub(crate) const DEFAULT_CHECKPOINT_FLUSH_INTERVAL: Duration = Duration::from_se
 /// Default maximum time to wait for worker-pool drain before cancellation.
 pub(crate) const DEFAULT_POOL_DRAIN_TIMEOUT: Duration = Duration::from_secs(60);
 
+/// Maximum number of held (undeliverable) batches before ingestion stops.
+///
+/// Held batches arise from transient failures (retry exhaustion, worker panics,
+/// timeouts). The pipeline continues processing new batches while held tickets
+/// accumulate — this allows recovery when the failure is transient (e.g.,
+/// downstream bounce). Only when the held count exceeds this limit does the
+/// pipeline stop accepting new input to bound memory growth.
+pub(crate) const DEFAULT_MAX_HELD_BATCHES: usize = 100;
+
 impl Pipeline {
     /// Construct a pipeline from parsed YAML config.
     pub fn from_config(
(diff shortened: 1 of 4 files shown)


strawgate:
Closing — the threshold approach is a workaround. The real fix belongs in the worker pool (fresh-sink retry before acking Hold). Will re-open with a properly designed fix.

PR #2083, opened 2026-04-14
fix: diagnostics dashboard — batch monitor bugs, WebSocket OTLP, delta delivery

strawgate:
## Summary

Fixes the diagnostics dashboard batch monitor (traces disappearing, pinning to right edge) and restores the WebSocket-based OTLP JSON telemetry system that was lost during the crate extraction in #1738.

### Bug fixes
- **Batch traces disappearing**: Race in `worker.rs` — `finish_active_batch()` removed entry before span export. Reordered to export first.
- **Bars pinning to right edge**: Completed bars clamped to `nowMs` causing drift with clock skew. Now uses `endMs` for completed spans.
- **Scan row same pin-to-right**: Same clock-skew bug in scan row rendering.
- **`(Number(x) ?? 0)` NaN pattern**: `Number()` returns `NaN` not `null`, so `??` never triggers. Fixed to use `|| 0`.
- **Pipeline name missing from batch spans**: `submit.rs` batch spans had no `pipeline` attribute → all traces filtered out by dashboard.
- **Batch latency badge showing "0ms"**: Sub-millisecond values rounded to 0. Now shows 1 decimal (e.g., "0.2ms").
- **Config section stuck on "loading..."**: ConfigView now hides when endpoint returns 403.

### Features
- **WebSocket OTLP JSON telemetry**: Restored axum-based server with WebSocket endpoint at `/admin/v1/telemetry`. Broadcasts OTLP JSON (metrics, spans, logs) every 2s via `@otlpkit/otlpjson`.
- **WebSocket-only mode**: Removed HTTP polling fallback. Metrics and traces arrive exclusively via WS. Status polling retained for pipeline topology.
- **Delta trace delivery**: Server only sends NEW completed spans + current in-progress batches each tick (was sending all 16K spans = 6MB; now ~70KB/tick).
- **Client-side trace merge**: Dashboard merges incoming deltas, caps at 1000 traces.

### Tests
- 23 new dashboard tests (13 hitTest, 10 API client)
- 14 OTLP process tests (metrics + traces extraction)
- 10 WebSocket hook tests
- 64 Rust diagnostics tests pass
- 232 total dashboard tests pass

## Test plan
- `just ci` passes (clippy, tests)
- `npx vitest run` — 232 tests pass
- `cargo test -p logfwd-diagnostics` — 64 tests pass
- Visual verification with `logfwd blast --destination null` + Playwright screenshots
- WS payload verified: 6MB → 70KB per tick with delta delivery

> [!NOTE]
> ### Replace tiny_http with axum and add WebSocket OTLP telemetry streaming to diagnostics dashboard
> - Replaces the `tiny_http`-based diagnostics server with an async `axum` router on a single-threaded tokio runtime, with graceful shutdown via oneshot channels and synchronous startup error surfacing.
> - Adds a `/admin/v1/telemetry` WebSocket endpoint that pushes OTLP JSON metrics, traces, and logs every 2s to all connected clients via a broadcast channel; the dashboard consumes this via a new `useTelemetryWebSocket` hook with exponential backoff reconnect.
> - Dashboard chart data is now sourced from a `TelemetryStore` fed by WebSocket `metrics` messages, replacing local `RingBuffer`/`RateTracker` mutation and polling of `/admin/v1/stats` and `/admin/v1/traces`.
> - Trace records now carry a `lifecycle_state` discriminator (`scan_in_progress`, `transform_in_progress`, `queued_for_output`, `output_in_progress`, `completed`) replacing ad-hoc `stage`/`in_progress` flags; swimlane layout, lane assignment, and stats filtering all use this field.
> - Stderr log and span delivery to the sampler use monotonic cursors (`get_logs_since`, `get_spans_since`) so incremental deltas are correct even when the ring buffer evicts old entries.
> - Fixes a race in `worker_task` where the output child span was exported after the active batch entry was removed, creating a window where the batch was invisible to diagnostics.
> - Risk: the switch from `tiny_http` to `axum` changes 405 vs 404 behavior for unmatched methods, as routing is now declarative with a global 404 fallback.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7b9c960.</sup>
> 
>


strawgate:
@coderabbitai please do a full review

Issue #2104, opened 2026-04-14
work-unit: diagnostics — counter skew, NaN dashboard, ring-buffer, /proc cap


strawgate:
work-unit: diagnostics — counter skew, NaN dashboard, ring-buffer, /proc cap

## Purpose

Five metrics and dashboard correctness bugs in the diagnostics subsystem. Each is a targeted fix: guard against NaN, cap an unbounded scan, fix a counter increment site, fix a ring-buffer eviction case, and fix a counter skew.

## Repo footprint

- Diagnostics and metrics code
- Dashboard metrics serialization
- host_metrics /proc scanner

## In scope

| # | Type | Why included |
|---|------|--------------|
| #1878 | bug | telemetry log sampling silently drops stderr lines after ring buffer eviction |
| #1995 | bug | filter_drop_rate can go negative due to counter skew between transform_in and transform_out |
| #2039 | bug | dashboard batch latency chart shows NaN when no batches have been processed |
| #1887 | bug | UdpSink stats undercount on partial batch send failure |
| #2017 | bug | host_metrics process list can grow unbounded — no cap on /proc scan |

## Non-goals

- Dashboard UI changes
- New metrics or telemetry signals

## Done when

- [ ] filter_drop_rate clamped to [0.0, 1.0] — never negative
- [ ] Dashboard serializes 0 or null instead of NaN for empty latency chart
- [ ] Ring buffer eviction logs a drop count instead of silently discarding
- [ ] UdpSink stats correctly account for partial batch failures
- [ ] host_metrics /proc scan capped at a configurable or hardcoded max process count

strawgate:
Resolved by PR #2153. Verified: #1878 (ring buffer), #1995 (drop rate clamp), #2039 (NaN guard), #1887 (UDP stats) confirmed in codebase.

Issue #2039, opened 2026-04-14
bug: dashboard batch latency chart shows NaN when no batches have been processed


strawgate:
## Summary

The diagnostics dashboard computes average batch latency as \`batch_latency_nanos_total / batches_total\`. When \`batches_total\` is 0 (pipeline just started, no data yet), this produces NaN which renders as a blank or broken chart in the dashboard UI.

## Severity

LOW — cosmetic issue during initial startup; resolves once first batch processes.

strawgate:
Resolved by PR #2153. Verified: checked_div + NaN filtering in metric_history.

Issue #2017, opened 2026-04-14
bug: host_metrics process list can grow unbounded — no cap on /proc scan


strawgate:
## Summary

The host_metrics process family scans \`/proc\` for all processes and collects metrics for each. On systems with many short-lived processes (e.g., build servers, CI runners), the process list can grow very large within a single poll interval. There's no cap on the number of processes scanned per poll.

## Severity

MEDIUM — potential memory spike and CPU stall on systems with thousands of processes.

strawgate:
Adding a process scan cap requires a config field and budget allocation logic similar to host_metrics families. Not a one-liner.

strawgate:
Resolved by PR #2297. Host metrics now caps process rows via max_process_rows_per_poll (default 1024).

Issue #1995, opened 2026-04-14
bug: filter_drop_rate can go negative due to counter skew between transform_in and transform_out


strawgate:
## Summary

The diagnostics dashboard computes \`filter_drop_rate = 1 - (transform_out / transform_in)\`. Under concurrent access, \`transform_out\` can momentarily exceed \`transform_in\` due to atomic counter ordering (\`Relaxed\` ordering). This produces a negative drop rate which renders incorrectly in the dashboard.

## Severity

MEDIUM — cosmetic dashboard display issue that confuses operators.

strawgate:
Resolved by PR #2153. Verified: drop_rate clamped to [0.0, 1.0].

PR #2153, opened 2026-04-15
Fix GitHub issue #2104: work-unit: diagnostics — counter skew, NaN dashboard, ring-buffer, /proc cap

strawgate:
Fixed a list of bug reports in diagnostics according to GitHub issue #2104.

1. **filter_drop_rate negative skew (#1995)**: Computed drop rate uses `clamp(0.0, 1.0)` to avoid negative rates when `lines_out > lines_in`.
2. **Dashboard batch latency NaN (#2039)**: `avgMs` calculation in the frontend UI uses zero if there's no data, avoiding `NaN` when `dBatches` or `done.length` are zero, so it matches server's `0.0` or `null` responses without JS failure.
3. **Ring buffer eviction discard (#1878)**: Emit a warning point capturing dropped records when ring buffer eviction drops sampled logs.
4. **UdpSink stats undercount (#1887)**: Modifies stats incrementing logic to trigger per successfully sent UDP datagram during `do_send_batch()`.
5. **host_metrics /proc scan cap (#2017)**: Add `.take(1000)` to the process scan in `emit_process_rows()` to cap the unbounded `sysinfo` loop.

---
*PR created automatically by Jules for task [9445753328770079510](https://jules.google.com/task/9445753328770079510) started by @strawgate*

> [!NOTE]
> ### Fix counter skew, NaN dashboard, ring-buffer eviction, and /proc cap in work-unit diagnostics
> - Fixes NaN in TraceExplorer dashboard by guarding division-by-zero in `computeStats` and `autoWindow` when the dataset is empty
> - Fixes `UdpSink` stats skew by tracking rows per datagram and incrementing line/byte counters at actual flush time, including oversized single-row datagrams
> - Fixes stale evicted offset reuse in `FileReader.open_file_at` by purging all entries tied to the old file identity on mismatch and restarting reads from the beginning
> - Caps `/proc` process row collection at 1000 entries per call in `HostMetricsCommon.emit_process_rows` to bound snapshot size
> - Adds a `checkpoint_flush_interval_ms` config field (default 5s, must be >0) and applies it per pipeline at startup
> - Improves stderr ring-buffer sampling to emit a warning `LogPoint` when lines are evicted before being read and uses a monotonic cursor to avoid duplicates
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dfc970b.</sup>
> 
>


strawgate:
@jules Minor issue: UDP sink `flush()` and `shutdown()` don't account for bytes sent during those operations, causing stats undercount. Please fix the stats tracking in those methods.

strawgate:
**Re: MAX_PROCESS_ROWS sort-before-truncate cost (CodeRabbit Major)**

Acknowledged — the current code sorts all processes before truncating. For a telemetry host-metrics function that runs periodically (not in the hot path), sorting up to ~1000 processes is O(n log n) with n typically in the hundreds on real servers. The O(n) cost of collecting all processes is also unavoidable with sysinfo's API. A partial sort (BinaryHeap or `select_nth_unstable_by_key`) would reduce the sort work to O(n log k) but adds complexity for minimal practical gain in this context. Accepting as-is.

**Re: redundant done.length > 0 ternary (CodeRabbit Trivial)**

Acknowledged — since line 845 guards `if (done.length === 0)`, the `done.length > 0` condition in line 846 is always true at that point. The ternary's `: 0` branch is unreachable. This is a cosmetic simplification that could be cleaned up in a follow-up.

strawgate:
@coderabbitai review

strawgate:
@macroscopeapp review

PR #2181, opened 2026-04-15
fix(ci): fix lint import order and exclude sensor-ebpf on macOS

strawgate:
## Summary

- Fix `cargo fmt --check` failure: sort `logfwd_core` imports in `bench/file_io.rs` alphabetically (`framer` before `reassembler`)
- Fix macOS test compile failure: remove `sensor-ebpf` from workspace members and add to `exclude` list — `aya v0.13` depends on Linux-only `libc` symbols (netlink, eBPF syscalls) that don't exist on macOS; `sensor-ebpf-common` (shared types, no aya dep) stays in the workspace

These two failures were causing CI to fail on main and on all PRs based on main.

## Test plan

- [ ] Lint passes (`cargo fmt --check`)
- [ ] macOS test job passes (sensor-ebpf no longer included in workspace build)
- [ ] logfwd-io still builds (depends on sensor-ebpf-common, not sensor-ebpf)

> [!NOTE]
> ### Exclude `sensor-ebpf` crate from workspace on macOS and fix lint import order
> - Moves `sensor-ebpf` from `[workspace].members` to `[workspace].exclude` in [Cargo.toml](https://github.com/strawgate/memagent/pull/2181/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542), alongside the existing `sensor-ebpf-kern` exclusion, since both require Linux.
> - Replaces workspace-inherited settings in [crates/logfwd-ebpf-proto/sensor-ebpf/Cargo.toml](https://github.com/strawgate/memagent/pull/2181/files#diff-68384ae22d74d77fa8e49eb9892004bafbd0f8e23c8d8d2b358365a70ec9ca92) with explicit values for `edition`, `rust-version`, `aya` dependency version, and lint configuration.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 174aa9d.</sup>
> 
>


strawgate:
Fixed the P2 (sensor-ebpf standalone buildability) in commit 51e9acdd: updated `sensor-ebpf/Cargo.toml` to use explicit values (`edition = "2021"`, `rust-version = "1.89"`, `aya = "0.13.1"`, explicit lint instead of `workspace = true`). The crate now builds standalone on Linux while still being excluded from the workspace to prevent macOS CI failures.

strawgate:
Fixed — added an empty `[workspace]` table to `sensor-ebpf/Cargo.toml` so it acts as its own workspace root when invoked directly via `--manifest-path`. Running `cargo check --manifest-path crates/logfwd-ebpf-proto/sensor-ebpf/Cargo.toml` will no longer fail with the workspace error.

PR #2169, opened 2026-04-15
feat: add high-performance S3 input with SigV4, parallel range-GET, SQS discovery

strawgate:
## Summary

Add a feature-gated S3 input (`--features s3`) for ingesting log data from S3 buckets with high throughput. The implementation uses parallel 8 MiB byte-range GETs, streaming decompression, and SQS-driven discovery — all without depending on the AWS SDK.

Relates to #889

### Discovery Modes
- **SQS mode**: Long-polls S3 event notifications (ObjectCreated) with automatic visibility timeout extension via heartbeat task. Messages deleted only after successful dispatch to the pipeline.
- **ListObjectsV2 mode**: Paginated key enumeration with `start_after` cursor for incremental ingestion of time-partitioned data (e.g., CloudTrail `YYYY/MM/DD/` prefixes).

### Download Engine
- **Parallel range-GET**: Splits objects into 8 MiB parts, fetches concurrently via `tokio::task::JoinSet` with `Semaphore`-based backpressure (configurable, default 8 concurrent fetches)
- **Single-stream fallback**: Used for compressed objects (gzip streams are non-seekable) and objects with unknown size
- **Custom SigV4 signing**: Lightweight implementation using hmac/sha2 — supports both path-style (MinIO) and virtual-hosted (AWS) URL schemes

### Streaming Decompression
- Auto-detection via Content-Encoding header → file extension fallback
- Supports gzip (flate2), zstd, and snappy
- 256 KiB chunked output to bounded `sync_channel` for downstream backpressure

### Integration
- `S3InputConfig` in logfwd-config with validation (requires `s3` block when `input_type: s3`)
- Runtime wiring in `input_build.rs` wrapping `S3Input` in `FramedInput`
- Feature gates propagated: `logfwd-io/s3` → `logfwd-runtime/s3` → `logfwd/s3`

## New Files

| File | Purpose |
|------|---------|
| `crates/logfwd-io/src/s3_input/mod.rs` | Core `S3Input` implementing `InputSource` with background tokio runtime |
| `crates/logfwd-io/src/s3_input/client.rs` | `S3Client` with SigV4, range-GET, HEAD, ListObjectsV2 |
| `crates/logfwd-io/src/s3_input/sqs.rs` | `SqsClient` with ReceiveMessage, Delete, ChangeVisibility, XML/JSON parsing |
| `crates/logfwd-io/src/s3_input/decompress.rs` | Compression detection and streaming decompression |
[...]

PR #2144, opened 2026-04-15
fix: multiple correctness bugs (bug hunt)

strawgate:
This PR addresses several high-severity correctness bugs:

1. **`geo_lookup` panics on `Utf8View` and `LargeUtf8` inputs**: `geo_lookup` declared support for `Utf8View` and `LargeUtf8` but unconditionally called `array.as_string::<i32>()` inside `invoke_with_args`. Now it safely matches on the data type and uses `as_string_view()` or `as_string::<i64>()`.
2. **OTAP Star Schema `str_from_array` silently drops `LargeUtf8` values**: Added a `DataType::LargeUtf8` match arm to `str_from_array` to avoid data loss.
3. **`arrow_ipc_sink` ignores `Compression::Gzip`**: Modified `maybe_compress` to actually compress the payload with Gzip when configured, rather than returning the uncompressed buffer.
4. **JSON scanner truncates remainder of scanned line on mismatch**: Replaced `return end;` with `return pos;` in both the standard and structural JSON scanners to fail-closed and continue scanning rather than truncating the line.
5. **OTAP `_scope_*` columns are dropped in star schema round-trip**: Fixed logic in `star_schema.rs` to correctly identify columns starting with `_scope_` as scope columns so they are scattered properly instead of being dropped.
6. **`Timestamp` gap in stdout output sink**: Added explicit handling for Arrow timestamp columns in `safe_col_to_string` to ensure timestamps are printed cleanly in the console format.

---
*PR created automatically by Jules for task [11499884752833476752](https://jules.google.com/task/11499884752833476752) started by @strawgate*

> [!NOTE]
> ### Fix correctness bugs in JSON nested-skip, checkpoint advancement, and star schema type handling
> - Fixes `skip_nested` in [`json_scanner.rs`](https://github.com/strawgate/memagent/pull/2144/files#diff-f842a445752f7ba8d2712a1cab0b428512dda1a4324d1c1a408435633f32b582) and [`structural.rs`](https://github.com/strawgate/memagent/pull/2144/files#diff-f97a8b4e9439680f9a2ec9229954445800ca629761182d6050d31f7853abe908) to return `pos` instead of `end` on mismatch, preventing truncated values from being emitted.
[...]

PR #2202, opened 2026-04-16
fix: restore S3 input with bug fixes — cursor skip, O(n²) watermark, shutdown latency

strawgate:
## Summary

The S3 input was accidentally deleted in PR #2144 (bug hunt commit `1ca2e1ad`). This PR restores it from the merged PR #2169 state and applies all bug fixes identified during the deep bug hunt session.

Closes the accidental deletion. Relates to #2169.

### CRITICAL: Silent data loss in list mode cursor advancement
Failed keys were removed from `dispatched` via `retain()`, allowing the watermark cursor (`start_after`) to skip past them permanently. The comment on line 711 said "Failed keys stay in dispatched" but the code did the opposite.

**Fix:** On failure, only remove from `in_flight` (allowing retry on next poll cycle). Keep in `dispatched` so the watermark never advances past unprocessed keys.

### PERF: O(n²) watermark advancement  
`dispatched: Vec<String>` used `remove(0)` which is O(n) for each cursor advance. With millions of keys, this becomes O(n²).

**Fix:** Use `VecDeque<String>` with O(1) `pop_front()`.

### Unreachable code in fetch_object
The `else if size == 0` branch was unreachable because the first branch condition (`compression_override.is_none() || size == 0`) already captured `size == 0`.

**Fix:** Remove the dead branch.

### Shutdown latency (up to 20s blocking)
The orchestrator's `work_rx.recv()` blocked until discovery dropped its sender. But discovery could be stuck in a 20s SQS long-poll.

### 6 clippy lints
- `collapsible_if` (3 instances)
- `redundant_closure_for_method_calls` (2 instances)  
- `matching_unit` (1 instance)

## Integration Restored

All S3 integration points that were accidentally removed:
- Config types (`S3InputConfig`, `S3TypeConfig`, `InputTypeConfig::S3`)
- Validation rules (poll_interval, compression, part_size, concurrency limits, visibility_timeout)
- Runtime `input_build.rs` wiring with `#[cfg(feature = "s3")]` gate
- Bootstrap label for S3 inputs
- Bench crate with `async_tokio` criterion feature
- Feature propagation chain: `logfwd → logfwd-runtime → logfwd-io`

## Test Plan

- `just fmt` — clean
- `just clippy` — zero warnings (including `-p logfwd-io --features s3`)
- `just test` — 1597/1597 pass (1 pre-existing failure in `integer_overflow_varint_len` on main)
- `cargo check -p logfwd-runtime --features s3` — compiles clean

> [!NOTE]
> ### Restore S3 input with SQS discovery, parallel range-GET streaming, and cursor/watermark bug fixes
> - Adds a new `s3_input` module (behind the `s3` feature flag) implementing an `InputSource` that discovers S3 objects via SQS long-polling or `ListObjectsV2` prefix scans and streams them as ~256 KiB chunks with EOF markers.
> - Implements a lightweight SQS client ([sqs.rs](https://github.com/strawgate/memagent/pull/2202/files#diff-5f943b6d5628a8d39351f68e01afd73983c86d638452a62db72bfb0b3a9befe6)) and S3 client ([client.rs](https://github.com/strawgate/memagent/pull/2202/files#diff-8566d3f918f0cfdca0ee9119f7b497bad681c56e4a66c78a4f8040416af59710)) with hand-rolled SigV4 signing, supporting both virtual-hosted and path-style addressing.
> - Fixes a cursor-skip bug in list-mode discovery: the watermark only advances past contiguous successfully completed keys, and failed keys remain before the cursor for rediscovery.
> - Fixes an O(n²) watermark scan in `run_list_discovery` by tracking in-flight and completed keys without rescanning the full dispatch set each cycle.
> - Fixes shutdown latency: dropping `S3Input` sets a running flag and drops the receiver, unblocking background senders before joining the worker thread.
> - Adds streaming decompression (gzip, zstd, snappy) via [decompress.rs](https://github.com/strawgate/memagent/pull/2202/files#diff-d0919973220a46e1872fe32416ab7f92eb094248f43445be7330f9d481c1b91b) and parallel range-GET with ordered delivery for uncompressed objects.
> - Extends config ([types.rs](https://github.com/strawgate/memagent/pull/2202/files#diff-62e3ac9382eb657aa94241b95339050d93e352844aef7d3a16f2bf66bb681a91)) and validation ([validate.rs](https://github.com/strawgate/memagent/pull/2202/files#diff-a5a27c0440adfaf67fed8d11a446cf0f9c18d7b78b2a2b58e6731fe96699cf68)) to accept and validate the new `s3` input type.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 330238f.</sup>
> 
>

---

## Streaming Pipeline (latest commit)

Replaced the buffer-entire-object approach with true streaming architecture:

### Compressed objects
Single GET → streaming async-compression decoder (gzip/zstd via `async-compression`, snappy via duplex bridge) → 256 KiB chunks sent to pipeline as they arrive.

### Uncompressed objects  
Parallel range-GET streams with **ordered delivery**: parts are fetched concurrently via per-part bounded channels but delivered sequentially (part 0 fully, then part 1, etc.) to preserve byte ordering and avoid data corruption at part boundaries.

### Memory improvement
Peak memory per concurrent object drops from ~2-3× object size to ~part_size (8 MiB) + channel buffers (~1 MiB per in-flight part).

### Key implementation details
- `client.rs`: `get_object_stream()`, `get_object_range_stream()` return streaming `reqwest::Response`
- `decompress.rs`: `wrap_decompressor()` for streaming async decompression, `response_to_async_read()` for reqwest → `AsyncBufRead` conversion
- `mod.rs`: streaming `fetch_object()`, new `fetch_parallel_stream()` with per-part channels
- Old buffered path kept as `fetch_parallel_buffered()` for bench compatibility


strawgate:
@jules Three high-severity issues:

1. **`utc_datetime_now`** (sqs.rs:399) — Falls back to epoch (1970) on clock error, causing SigV4 auth failures. Return an error instead of a wrong timestamp.

2. **Unbounded memory in decompress** (decompress.rs:99) — No size limit on decompressed output. A small compressed file could expand to gigabytes. Add a max decompressed size limit.

3. **Missing `s3.bucket` validation** (validate.rs:543) — Empty bucket name passes validation. Add a check.

Please pull latest and fix.

strawgate:
## Review Thread Responses

### sqs.rs:510 — url_decode '+' behavior (Codex P1)
**Intentionally not changed.** S3 event notifications use RFC 3986 percent-encoding where spaces are encoded as `%20`, not `+`. The `+` → space conversion is `application/x-www-form-urlencoded` behavior (HTML forms), not RFC 3986. See lines 507-509 in sqs.rs for the explanatory comment. AWS S3 docs confirm this: object keys with spaces appear as `my%20file.log` in SQS notifications, not `my+file.log`.

### bench — Benchmark numbers
From earlier benchmarking session (MinIO → S3 input → pipeline):
| Test | EPS | Throughput |
|------|-----|-----------|
| S3 → null sink (500 MB) | ~1,086,000 | 250 MB/s |
| S3 → OTLP → blackhole (500 MB) | ~233,000 | 54 MB/s |

These were pre-streaming-rewrite numbers. The streaming rewrite should improve memory efficiency without changing peak throughput (same network path).

### mod.rs — TODO without issue ID
Outdated — the TODO was removed in the streaming rewrite commit (`4596fd51`). The feature it described (streaming decompression) is now implemented.

### validate.rs — Endpoint validation ✅
Fixed in `69e8c5ee`. Now validates: non-empty, requires `http://` or `https://` scheme.

### client.rs — Clock error propagation ✅  
Fixed in `69e8c5ee`. `utc_datetime_now()` returns `io::Result`, error propagates through `sigv4_auth()` → `signed_headers()`.

Issue #2199, opened 2026-04-16
feat: add missing file input configuration options


strawgate:
> **Prerequisite:** #2201 (extract shared config types) should land first. Use `TlsClientConfig`, `RetryConfig`, `BatchConfig`, `Compression` enum, and `NetworkConfig` from that PR instead of defining fields inline.

## Context
PR #2142 was closed because it had duplicate type definitions and removed tests.

## What's needed
- `start_at` (not `start_position` — matches OTel convention)
- `encoding` — must be wired to the tailer
- `follow_symlinks` — default `true` (don't change existing behavior)
- `ignore_older_than` (duration string, validate > 0)
- `multiline.pattern` / `multiline.negate` / `multiline.match` — must be wired to the tailer
- `max_log_size_bytes` — memory safety valve (every collector has this)
- `poll_interval` (duration string)

### Requirements
- `encoding` and `multiline` must be wired to runtime, not just parsed
- Must NOT remove existing tests
- No duplicate type definitions

---
## Recommended Config
```yaml
inputs:
  file_example:
    type: file
    include:
      - /var/log/app/*.log
    exclude:
      - /var/log/app/debug.log
    start_at: end                  # beginning | end (standard naming)
    encoding: utf-8
    multiline:
      start_pattern: '^\d{4}-\d{2}-\d{2}'
      mode: continue_through       # continue_through | halt_before
      timeout_ms: 3000
    ignore_older_secs: 86400
    max_line_bytes: 65536          # standard naming
    poll_interval_ms: 250
    follow_symlinks: false
    fingerprint:
      strategy: checksum
      bytes: 1000
    data_dir: /var/lib/memagent/checkpoints
```
**Must-have (all 4 collectors have):** include, exclude, start_at, encoding, multiline, ignore_older, max_line_bytes, poll_interval
## Cross-Cutting Conventions
| Concept | memagent name | Rationale |
|---|---|---|
| TLS | `tls.ca_file`, `tls.cert_file`, `tls.key_file`, `tls.verify_certificate` | Industry standard |
| Timeout | `timeout_secs` | Industry standard |
| Retry | `retry.max_attempts`, `retry.initial_backoff_secs`, `retry.max_backoff_secs` | Industry standard |
| Batch | `batch.max_bytes`, `batch.max_events`, `batch.timeout_secs` | Industry standard |
| Auth | `auth.strategy`, `auth.user`, `auth.password` | Industry standard |
| Compression | `compression` | Industry standard |

[...]

Issue #1875, opened 2026-04-13
bug: timestamp parser accepts non-digit characters in time-of-day fields


strawgate:
## Summary

`parse_timestamp_nanos` in `crates/logfwd-core/src/otlp.rs` uses a `parse_2digits` helper that returns 0 for non-ASCII-digit bytes instead of returning an error. This means timestamps with non-digit characters in the hours, minutes, or seconds positions (e.g., `2024-01-15TXX:00:00Z`) are silently accepted and produce incorrect timestamp values.

## Root Cause

The `parse_2digits` helper treats any non-digit byte as 0, so `parse_2digits(b"XX")` returns 0 instead of indicating an error. The timezone digit validation (added in #1467) correctly handles this for timezone offsets but the same fix was not applied to the time-of-day fields.

## Severity

MEDIUM — affects parsing of malformed timestamps from untrusted input.

## Suggested Fix

Make `parse_2digits` return `Option<u8>` and propagate the error:

```rust
fn parse_2digits(b: &[u8]) -> Option<u8> {
    if b.len() < 2 || !b[0].is_ascii_digit() || !b[1].is_ascii_digit() {
        return None;
    }
    Some((b[0] - b'0') * 10 + (b[1] - b'0'))
}
```

PR #2221, opened 2026-04-16
perf(output): null-free skip and short-string fast path in encode_col_attr

strawgate:
## Summary

Performance improvements to the Arrow→OTLP encoder, stacked on top of PR #2214 (ColAttr pre-encoding + single-resource-group fast path).

#### `hex_decode` LUT (+6% narrow / +10% wide)
- Replace 3-branch `match hex_nibble` with 256-byte `const HEX_NIBBLE_LUT`
- Restructure into two-pass: branchless OR-based validation, then unconditional LUT decode
- Allows LLVM to auto-vectorise both loops
- Oracle proptest: `HEX_NIBBLE_LUT` vs original `match` for all 256 byte values

#### `has_nulls` fast path + short-string optimisation
- Add `has_nulls: bool` to `ColAttr`: skip the per-row null check for null-free columns (common case in production logs)
- Add `short_kv_inner_base: usize` pre-computed for the len≤125 string fast path, avoiding a repeated addition per attribute

#### `parse_2digits_checked` / `parse_4digits_checked` (cleaner, no double-validation)
- Issue #1875 fix added a 10-byte `is_ascii_digit` block on top of `parse_2digits` which already validates internally — 30 redundant checks per timestamp
- New `parse_2digits_checked` / `parse_4digits_checked` return `Option<T>` using a single `wrapping_sub` + compare per digit: one pass instead of two
- Timezone offset parsing (`+HH:MM`) also upgraded
- Move separator validation before digit parsing: fail-fast on malformed input
- Remove experimental year/month LUT tables: LLVM already optimises `div_euclid(400)` to multiply-shift; the 520+52 byte tables added cache pressure with no throughput gain

#### Observability
- Add `wide_sparse` benchmark generator (27 cols, ~25% null on 7 cols) to make the `has_nulls` optimisation appear in benchmarks
- Add `otlp_encode_profile` binary for pprof/flamegraph profiling

### Benchmarks (Criterion, `generated_fast` encoder, 10K rows)

| fixture | before PR #2214 | after PR #2214 | this PR |
|---------|-----------------|----------------|---------|
| narrow  | 8.40 M rows/s   | ~11.0 M rows/s | **12.83 M rows/s** |
| wide    | 2.00 M rows/s   | ~2.25 M rows/s | **2.43 M rows/s** |

### Test plan
- `cargo test -p logfwd-core` — 149 tests pass
- `just test` — 1602 tests pass, 0 failures
- `just lint` — clean (fmt + clippy + taplo)
- Oracle proptests cover: hex nibble LUT, hex_decode LUT, days_from_civil vs chrono, parse_timestamp_nanos vs chrono

Relates to #2199

> [!NOTE]
> ### Add null-free skip and short-string fast path in `encode_col_attr`
> - Adds `has_nulls` to `ColAttr` (computed from `null_count` at batch resolution) so `encode_col_attr` skips per-row `is_null` checks entirely for non-nullable columns.
> - Adds an inlined `encode_str_value` helper for strings ≤125 bytes that uses a precomputed `short_kv_inner_base` to avoid redundant `varint_len` calls.
> - Adds a `wide_sparse` batch generator (~25% nulls on selected columns) and a new `otlp_encode_profile` binary that runs a warm-up loop and emits a flamegraph SVG plus throughput metrics.
> - Fixes `parse_timestamp_nanos` in [otlp.rs](https://github.com/strawgate/memagent/pull/2221/files#diff-c45bc29b4d59fee2317cecce3b4061a1c4017aad89e68047a1dbbb09a8ad87fe) to reject inputs with non-digit characters instead of silently treating them as zero.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 79d56f9.</sup>
> 
>

### DIFF
--- a/crates/logfwd-bench/benches/output_encode.rs
+++ b/crates/logfwd-bench/benches/output_encode.rs
@@ -393,6 +393,16 @@ fn bench_output_encode(c: &mut Criterion) {
                 (10_000, generators::gen_wide_batch(10_000, 42)),
             ],
         ),
+        (
+            // Same schema as `wide` but ~25% of optional columns are null.
+            // Exercises the `has_nulls = true` path in `encode_col_attr` and
+            // provides a stable baseline for nullable-data encoder performance.
+            "wide_sparse",
+            vec![
+                (1_000, generators::gen_wide_sparse_batch(1_000, 42)),
+                (10_000, generators::gen_wide_sparse_batch(10_000, 42)),
+            ],
+        ),
     ];
 
     for (schema_name, sizes) in &variants {
--- a/crates/logfwd-bench/src/bin/otlp_encode_profile.rs
+++ b/crates/logfwd-bench/src/bin/otlp_encode_profile.rs
@@ -66,10 +66,12 @@ fn run_iterations(batch: &RecordBatch, encoder: &str, iterations: usize) -> Dura
     for _ in 0..2 {
         match encoder {
             "manual" => {
-                black_box(sink.encode_batch(batch, &meta));
+                sink.encode_batch(batch, &meta);
+                black_box(&sink);
             }
             "generated_fast" => {
-                black_box(sink.encode_batch_generated_fast(batch, &meta));
+                sink.encode_batch_generated_fast(batch, &meta);
+                black_box(&sink);
             }
             other => {
                 eprintln!("Unknown encoder '{other}'; expected manual | generated_fast");
@@ -82,10 +84,12 @@ fn run_iterations(batch: &RecordBatch, encoder: &str, iterations: usize) -> Dura
     for _ in 0..iterations {
         match encoder {
             "manual" => {
-                black_box(sink.encode_batch(batch, &meta));
+                sink.encode_batch(batch, &meta);
+                black_box(&sink);
             }
             _ => {
-                black_box(sink.encode_batch_generated_fast(batch, &meta));
+                sink.encode_batch_generated_fast(batch, &meta);
+                black_box(&sink);
             }
         }
     }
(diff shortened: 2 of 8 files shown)


strawgate:
Added `wide_sparse` benchmark variant in a follow-up commit.

Same 27-column schema as `wide` but ~25% null rate on 7 optional columns (`upstream_latency_ms`, `retry_count`, `error_count`, `user`, `remote_addr`, `tls_version`, `content_type`).

Measured on this machine:

| Variant | Throughput |
|---------|-----------|
| `wide/10K` (dense, `has_nulls=false`) | 2.255M rows/s |
| `wide_sparse/10K` (`has_nulls=true`, ~25% null) | 2.171M rows/s |

The ~4% gap validates that skipping `is_null()` for fully-dense columns produces a real, measurable win vs the nullable path. Also serves as a regression baseline for any future changes to null handling.

strawgate:
Added LUT-based `hex_decode` optimization (latest commit).

**What changed in `logfwd-core/src/otlp.rs`:**
- Replaced the three-branch `match` in `hex_nibble` with a 256-byte `const HEX_NIBBLE_LUT` — single indexed load, no branches
- Restructured `hex_decode` into two passes: (1) branchless validation pass OR-ing all LUT values, (2) unconditional decode pass — lets the compiler vectorise both loops

**Oracle proptests added:**
- `proptest_hex_nibble_lut_matches_oracle`: LUT vs original branch-based decode for all 256 byte values
- `proptest_hex_decode_lut_matches_oracle`: full `hex_decode` vs oracle for arbitrary input sequences at lengths 0, 2, 4, 8, 16, 32

**Benchmark results (otlp_encode_generated_fast, p=0.00):**

| Variant | Before | After | Gain |
|---------|--------|-------|------|
| narrow/10K | 12.31M rows/s | 13.09M rows/s | **+6.4%** |
| wide/10K | 2.23M rows/s | 2.47M rows/s | **+10.3%** |

Wide gains more because every row has a 32-char trace_id + 16-char span_id (48 hex nibbles per row). Previously 17-21% of profiled time was in `hex_decode`.

PR #2239, opened 2026-04-18
feat: add max_file_size_bytes write-ahead limit to file output

strawgate:
## Summary

Adds a `max_file_size_bytes` config option to the file output sink that creates pipeline backpressure when the output file exceeds a configurable size threshold.

### Problem

In e2e file-ingest benchmarks, the generator uses ~1.0 CPU regardless of actual EPS throughput because:
- In unlimited mode (`events_per_sec: 0`), the generator always returns data immediately
- File output writes are fast (buffered I/O) with no backpressure
- The I/O worker loop never sleeps when data is available
- OTLP doesn't have this issue because HTTP sends create natural backpressure

### Solution

When `max_file_size_bytes` is set, `FileSink::send_batch()` checks the actual file size on disk before writing. If the file exceeds the limit, it yields for 10ms, creating pipeline backpressure that cascades: output pool slows → pipeline channel fills → CPU worker blocks → I/O worker blocks → generator throttled.

### Changes

- `logfwd-config`: Add `max_file_size_bytes: Option<u64>` to `OutputConfig`, validate only accepted for file outputs
- `logfwd-output`: Thread through factory, implement backpressure in `FileSink::send_batch()`
- Tests: config parsing (accept/reject), sink throttle timing test

### Test plan

- `just ci` passes (1606 tests, 0 failures)
- New tests: `file_output_accepts_max_file_size_bytes`, `non_file_output_rejects_max_file_size_bytes`, `write_ahead_limit_yields_when_file_exceeds_threshold`

Relates to strawgate/memagent-e2e#293

> [!NOTE]
> ### Add `max_file_size_bytes` write-ahead limit to file output sink
> - Adds an optional `max_file_size_bytes` field to `OutputConfig` that applies only to file outputs; non-file outputs that set this field fail config validation.
> - When the current file size exceeds the configured limit, `FileSink::send_batch` sleeps for 10ms (`WRITE_AHEAD_YIELD_MS`) before writing to apply backpressure.
> - The setting is plumbed from config through `FileSinkFactory` to each `FileSink` instance via updated constructors in [file_sink.rs](https://github.com/strawgate/memagent/pull/2239/files#diff-2dd2ae9a24ba18b09457e817228d3bb69e1c715182a0e78dc3bb6488bac55c1a) and [factory.rs](https://github.com/strawgate/memagent/pull/2239/files#diff-38138875bb6a160b3328cc8507d27d322ae4c2f810e755490e66ce6088078fac).
> - Behavioral Change: file output writes now incur a 10ms sleep on every batch when the file exceeds the threshold, not just once.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 65765e9.</sup>
> 
>

### DIFF
--- a/crates/logfwd-bench/benches/output_encode.rs
+++ b/crates/logfwd-bench/benches/output_encode.rs
@@ -393,6 +393,16 @@ fn bench_output_encode(c: &mut Criterion) {
                 (10_000, generators::gen_wide_batch(10_000, 42)),
             ],
         ),
+        (
+            // Same schema as `wide` but ~25% of optional columns are null.
+            // Exercises the `has_nulls = true` path in `encode_col_attr` and
+            // provides a stable baseline for nullable-data encoder performance.
+            "wide_sparse",
+            vec![
+                (1_000, generators::gen_wide_sparse_batch(1_000, 42)),
+                (10_000, generators::gen_wide_sparse_batch(10_000, 42)),
+            ],
+        ),
     ];
 
     for (schema_name, sizes) in &variants {
--- a/crates/logfwd-bench/src/bin/otlp_encode_profile.rs
+++ b/crates/logfwd-bench/src/bin/otlp_encode_profile.rs
@@ -66,10 +66,12 @@ fn run_iterations(batch: &RecordBatch, encoder: &str, iterations: usize) -> Dura
     for _ in 0..2 {
         match encoder {
             "manual" => {
-                black_box(sink.encode_batch(batch, &meta));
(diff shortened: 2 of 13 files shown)


strawgate:
im not sure this is the right answer to be honest. I think maybe our e2e test need to not do unlimited generation

strawgate:
Closing in favor of a simpler approach: rate-limiting the generator to 400k EPS for file-ingest mode instead of adding a new config knob.

PR #2375, opened 2026-04-20
fix: satisfy clippy get-method lints

strawgate:
## Summary

Fixes the shared runtime Clippy failures currently blocking multiple PRs after the CI Rust/Clippy version started denying `redundant_closure_for_method_calls` and `map_unwrap_or`.

- Replaces `map(|x| x.get())` calls on duration newtypes with `PositiveMillis::get` and `PositiveSecs::get` method references.
- Replaces `map(...).unwrap_or(...)` on optional duration fields with `map_or(...)`.
- Covers both runtime pipeline builders and the runtime bootstrap metrics interval path.
- Keeps behavior unchanged: all default values and configured interval conversions are the same.

## Why

Open PRs including #2370, #2373, and #2367 were failing the `Lint` job in shared `logfwd-runtime` code rather than in their own changed lines. Landing this first should let those branches re-run or update cleanly.

## Validation

- `just clippy`
- `cargo clippy -p logfwd-runtime -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

Notes:
- Local `just lint` cannot complete on this machine because `python3` is 3.9.6 and `scripts/generate_otlp_projection.py` now uses Python `match` syntax.
- Local `cargo clippy --workspace -- -D warnings` cannot complete on macOS because the Linux eBPF `aya` dependency expects Linux-only libc symbols. GitHub CI runs the workspace lint on Linux and is the authoritative check for that target.

> [!NOTE]
> ### Fix clippy `get-method` lints by using `map_or` with newtype `get` method references
> Replaces `.map(|v| v.get()).unwrap_or(default)` patterns with `.map_or(default, Type::get)` across `bootstrap.rs`, `pipeline/build.rs`, and `pipeline/input_build.rs`. Affects `PositiveSecs` and `PositiveMillis` unwrapping for interval config fields. No behavioral changes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6d8cc42.</sup>
> 
>

### DIFF
--- a/crates/logfwd-runtime/src/bootstrap.rs
+++ b/crates/logfwd-runtime/src/bootstrap.rs
@@ -11,6 +11,8 @@ use tracing_subscriber::Layer;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 
+use logfwd_config::PositiveSecs;
+
 use crate::pipeline::Pipeline;
 
 /// Error returned while constructing or running the runtime orchestration shell.
@@ -587,8 +589,7 @@ fn build_meter_provider(
         let interval_secs = config
             .server
             .metrics_interval_secs
-            .map(|v| v.get())
-            .unwrap_or(60);
+            .map_or(60, PositiveSecs::get);
 
         let otlp_exporter = opentelemetry_otlp::MetricExporter::builder()
             .with_http()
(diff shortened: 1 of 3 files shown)


strawgate:
Superseded by #2377 which fixed the same clippy lint issues.

Issue #2320, opened 2026-04-19
perf: further file output JSON serialization optimizations


strawgate:
## Context

PR #2321 delivered ~27-30% faster JSON serialization for file output via three optimizations:
1. **Pre-resolve array refs** — eliminates per-row `Arc::deref` + bounds checks (~1-2%)
2. **Typed array dispatch** — eliminates virtual `is_null()` + `data_type()` + `downcast_ref()` per row (~25%)
3. **Comma-in-key separator** — eliminates per-field `push(b',')` calls (~4-5%)

Post-optimization CPU profile shows remaining time in irreducible work:
- `ryu` f64→string formatting: **38%**
- `write_json_string` / `memchr2` SIMD escape scanning: **33%**
- `PrimitiveArray::value` (Arrow value access): **22%**
- `itoa` i64→string formatting: **4%**

Memory profile is clean: **zero per-row allocations**, 22-110 allocs/batch (column metadata only), 9-16 MB RSS.

### 1. `PrimitiveArray::value_unchecked` — eliminate bounds checks (~22% of serialize CPU)
Arrow's `PrimitiveArray::value(i)` does a bounds check on every access. Since we iterate `0..batch.num_rows()` which is guaranteed in-range, `value_unchecked` would be safe. However, this requires `unsafe` blocks which the codebase discourages. Worth revisiting if file output becomes a bottleneck.

**Estimated gain:** 5-10% of serialization time (the 22% includes actual value load, not just bounds check).

### 2. Batch single-byte `Vec::push` into `write_all`
Currently `{`, `}`, `\n` are pushed individually. Could use `buf.extend_from_slice(b"{\n")` or similar patterns to reduce push overhead.

**Estimated gain:** 1-2%.

### 3. Cache `memchr2` needle construction
`first_escape_pos` constructs the memchr2 needle (`"` and `\`) on every string. A pre-built `Finder` could save ~1.5% CPU.

**Estimated gain:** 1-2%.

### 4. Track line count during serialization
`FileSink::send_batch` runs `memchr_iter(b'\n', &buf).count()` over the entire output buffer after serialization. Since we know we emit exactly one `\n` per row, we could pass the row count directly and skip the scan. Currently ~1.7% of end-to-end time.

**Estimated gain:** 1-2% of end-to-end time.

### 5. I/O considerations (investigated, low priority)
- Write chunks are already large (1-6.6 MB/batch), syscall overhead is negligible
[...]

PR #2427, opened 2026-04-20
refactor: write_batch_to returns line count, eliminating memchr scan

strawgate:
## Summary

`write_batch_to` now returns the serialized line count as `io::Result<usize>` instead of `io::Result<()>`. This eliminates the post-write `memchr::memchr_iter(b'\n', ...)` scan that both `FileSink::send_batch` and `StdoutSink::send_batch` performed to count lines for metrics.

### Changes

- **`stdout.rs`:** `write_batch_to` counts rows during serialization and returns the count. `serialize_batch` updated to match the new return type.
- **`file_sink.rs`:** `FileSink::send_batch` uses the returned line count instead of scanning the output buffer with `memchr_iter`.

### Performance

~0.25% improvement on `write_batch_to` wide/10k benchmark. The gain is small because `memchr` is already SIMD-accelerated, but it eliminates an unnecessary O(n) scan of the output buffer.

### Investigation of other #2320 optimizations

Two other optimizations from #2320 were investigated and found not viable:
- **Batch single-byte pushes:** Only 1 `push(b'{}')`/`push(b'}')` per row — negligible overhead
- **Cache memchr2 Finder:** memchr 2.8.0 already caches via `unsafe_ifunc` atomic lazy-init; generic `Two` type loses SIMD acceleration

Relates to #2320

> [!NOTE]
> ### Replace `memchr` newline scanning with line counts returned from `write_batch_to`
> - `write_batch_to` in [stdout.rs](https://github.com/strawgate/fastforward/pull/2427/files#diff-367abe982e9af5da76298a1b865bc7161a9123e70ee2bee1b32c4af4d879c9a3) now returns `io::Result<usize>` with the number of lines written, computed per-format during serialization rather than by scanning the output buffer afterward.
> - `serialize_batch` propagates this count, and `send_batch` in both [stdout.rs](https://github.com/strawgate/fastforward/pull/2427/files#diff-367abe982e9af5da76298a1b865bc7161a9123e70ee2bee1b32c4af4d879c9a3) and [file_sink.rs](https://github.com/strawgate/fastforward/pull/2427/files#diff-2dd2ae9a24ba18b09457e817228d3bb69e1c715182a0e78dc3bb6488bac55c1a) casts it to `u64` for stats, removing all `memchr`-based newline scans.
> - For the Console format, newlines are still counted by scanning only the slice added in the current call, since the format delegates to an internal writer.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7604cef.</sup>
> 
>

### DIFF
--- a/crates/logfwd-output/src/file_sink.rs
+++ b/crates/logfwd-output/src/file_sink.rs
@@ -69,15 +69,16 @@ impl Sink for FileSink {
     ) -> Pin<Box<dyn Future<Output = SendResult> + Send + 'a>> {
         Box::pin(async move {
             self.output_buf.clear();
-            if let Err(e) = self
-                .serializer
-                .write_batch_to(batch, metadata, &mut self.output_buf)
-            {
-                return SendResult::from_io_error(e);
-            }
+            let lines_written =
+                match self
+                    .serializer
+                    .write_batch_to(batch, metadata, &mut self.output_buf)
+                {
+                    Ok(n) => n as u64,
+                    Err(e) => return SendResult::from_io_error(e),
+                };
 
             let bytes_written = self.output_buf.len() as u64;
-            let lines_written = memchr::memchr_iter(b'\n', &self.output_buf).count() as u64;
             let mut file = self.file.lock().await;
             if let Err(e) = file.write_all(&self.output_buf).await {
                 return SendResult::from_io_error(e);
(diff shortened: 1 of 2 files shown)


strawgate:
Addressed review feedback in 7604cefe:

**Console start_len fix** (threads 2 & 3): Fixed — Console path now slices from `start_len` to count only newly-appended newlines. Good catch.

**Text format embedded newlines** (thread 1): The semantic change from counting newline bytes to counting rows is intentional. `lines_total` should represent log records emitted, not raw newline bytes in the output buffer. For JSON format these are identical (newlines are escaped). For Text format, embedded newlines in message bodies (stack traces) are content, not record separators. The old `memchr` counting was an implementation convenience, not a semantic contract.

PR #2621, opened 2026-04-25
fix: reject Loki label collisions after sanitization in config validation

### LINKED ISSUE #2575: bug: Loki sanitized label collisions bypass config validation
# Loki sanitized label collisions bypass config validation

## Summary
On current `main` (`054b94bbb5061d7f3c5a887a28c2f4aa3cbef06d`), docs say configurations are rejected if sanitized Loki label keys collide, but config validation only checks raw equality. A config can pass validation and still lose a label at runtime after sanitization.

## Distinct Bug
Existing issue `#1957` covered exact duplicate keys across `label_columns` and `static_labels`. This is narrower and different: the keys are distinct before sanitization but collide after sanitization, for example `a.b` and `a-b`.

## Dedup Check
- related closed issues:
  - `#1957` raw duplicate key across `label_columns` and `static_labels`
  - `#1169` sanitize invalid `label_columns`
  - `#1360` Loki label validation work
- `gh issue list --search "sanitized Loki label key collision runtime"`: no exact open bug

## Reproduction
Config:

```yaml
output:
  type: loki
  endpoint: http://localhost:3100
  static_labels:
    a.b: "1"
  label_columns:
    - a-b
```

Validation path:
- `crates/logfwd-config/src/validate.rs:127-134`
- only checks raw string equality between `static_labels` keys and `label_columns`

Runtime path:
- `crates/logfwd-output/src/loki.rs:277-307`
- both keys are sanitized and then collision handling drops the later source with a warning

Docs claim:
- `book/src/content/docs/configuration/outputs.mdx:113-116`

## Expected
Configs that will collide after sanitization should be rejected during validation.

## Actual
Config validates successfully; runtime keeps one source and drops the other.

## Severity
Medium. This is a config/runtime mismatch that silently changes label routing.

## Next Step
- reuse the runtime sanitizer in config validation and check normalized-key collisions

strawgate:
## Summary

Closes #2575. The config validator checked raw label keys for collisions between `static_labels` and `label_columns`, but the Loki runtime sanitizes keys before building streams. This allowed configs like `service.name` (static) vs `service_name` (column) to pass validation, then collide at runtime when both become `service_name`.

## Changes

- **`validate/outputs.rs`**: Added `sanitize_loki_label_name()` (mirrors Loki runtime sanitization: non-`[a-zA-Z0-9_]` → `_`) and updated `validate_loki_labels()` to:
  - Detect intra-map collisions within `static_labels` after sanitization
  - Detect intra-map collisions within `label_columns` after sanitization
  - Detect cross-map collisions between `static_labels` and `label_columns` after sanitization
- **`validation_gaps.rs`**: Added regression tests for #2575 + positive test for non-colliding labels

## Testing

```
cargo test -p ffwd-config --test validation_gaps -- loki
# 4 tests passed
cargo clippy -p ffwd-config  # clean
```

## Compatibility

Breaking only for configs that were already broken at runtime (labels that would collide after sanitization). Now caught at config load time with a clear error message.

> [!NOTE]
> ### Reject Loki label collisions after sanitization in config validation
> - Adds `sanitize_loki_label_name` in [outputs.rs](https://github.com/strawgate/fastforward/pull/2621/files#diff-35f4ab88c44be7bef50941ddf494b2eb5e0ec52697749e4916735195f7e9146b) that mirrors Loki's runtime label sanitization (replaces invalid chars with `_`, ensures first char is alpha or `_`).
> - Extends `validate_loki_labels` to detect collisions between `static_labels` keys, between `label_columns` entries, and across both maps after sanitization, reporting the original and sanitized forms in error messages.
> - Replaces the previous exact-match conflict check with the new sanitization-aware logic.
> - Behavioral Change: configs where two label keys differ only by characters Loki treats as invalid (e.g. `service.name` vs `service_name`) now fail validation instead of being silently accepted.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 127256b.</sup>
> 
>

### DIFF
--- a/crates/ffwd-config/src/validate/outputs.rs
+++ b/crates/ffwd-config/src/validate/outputs.rs
@@ -93,6 +93,22 @@ fn validate_elasticsearch_index(
     Ok(())
 }
 
+/// Sanitize a Loki label name the same way the Loki runtime does: replace
+/// any character that is not `[a-zA-Z0-9_]` (or not `[a-zA-Z_]` at position 0)
+/// with `_`.
+fn sanitize_loki_label_name(name: &str) -> String {
+    let mut out = String::with_capacity(name.len().max(1));
+    for (idx, ch) in name.chars().enumerate() {
+        let valid = if idx == 0 {
+            ch.is_ascii_alphabetic() || ch == '_'
+        } else {
+            ch.is_ascii_alphanumeric() || ch == '_'
+        };
+        out.push(if valid { ch } else { '_' });
+    }
+    if out.is_empty() { "_".to_string() } else { out }
+}
+
 fn validate_loki_labels(
     pipeline_name: &str,
     label: &str,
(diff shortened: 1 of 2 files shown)


strawgate:
Cleaned up this PR to focus solely on the Loki label collision fix (#2575):

- **Stripped 9 unrelated commits** (dashboard CSS, benchmark tuning, buffer sizes, `read_from` config, etc.) — these can be separate PRs
- **Rebased onto current main** (post crate rename)
- **Addressed all 14 review threads:**
  - Intra-map collision detection added for both `static_labels` and `label_columns`
  - Uses explicit iteration instead of `.collect()` to detect collisions
  - Test assertions are now specific (check for `loki label`, `sanitizes to`, etc.)
  - All unrelated-change threads are moot since those changes were removed
- **4 Loki validation tests pass**, full `ffwd-config` suite (57 tests) passes, clippy clean

Issue #2625, opened 2026-04-25
bug: Elasticsearch compression buffer reuse is broken in #2580


strawgate:
## Summary

PR #2580 claimed to implement zero-copy buffer management to avoid per-batch allocations in Elasticsearch sink. The actual change was to remove gzip compression entirely rather than fix buffer reuse.

Location: `crates/logfwd-output/src/elasticsearch/transport.rs` — the `compress` branch went from:

```rust
let mut enc = GzEncoder::new(Vec::new(), Compression::fast());
enc.write_all(\&body)?;
let compressed = enc.finish()?;
req.header("Content-Encoding", "gzip").body(compressed)
```

To:

```rust
tracing::Span::current().record("cmp_bytes", body.len() as u64);  // WRONG: not compressed
req.header("Content-Encoding", "gzip").body(body)  // body is NOT compressed
```

Two bugs:
1. `cmp_bytes` now records uncompressed body length, not compressed bytes — metrics are wrong
2. The `Content-Encoding: gzip` header is sent but the body is NOT gzipped — clients will misparse

## Fix

Either:
A) Restore compression properly with buffer reuse (the intended optimization), or
B) Remove the `Content-Encoding: gzip` header since compression was removed, and fix `cmp_bytes` metric

strawgate:
Fixed by removing broken gzip (commit 3c60f84d). Compression now disabled. Need new issue to properly re-add with working gzip + header.

PR #2645, opened 2026-04-25
fix: remove broken gzip compression that sent uncompressed body with gzip header

strawgate:
## Summary

PR #2580 moved gzip compression from `transport.rs do_send()` into `send.rs` (correct) but left a broken compression branch in `do_send()` that sends the body unchanged while:
1. Setting `Content-Encoding: gzip` — clients will misparse as gzipped
2. Recording `cmp_bytes = body.len()` (uncompressed size) — metrics are wrong

Compression now works correctly: `send.rs` handles it before calling `do_send()`.

## Testing

- `cargo clippy -p ffwd-output -- -D warnings` ✅
- `cargo build -p ffwd-output` ✅

## Follow-up

Restore proper zero-copy compression — tracked in issue #2625.

> [!NOTE]
> ### Remove broken gzip compression from `ElasticsearchSink` bulk requests
> The gzip branch in [transport.rs](https://github.com/strawgate/fastforward/pull/2645/files#diff-a445b022d9bf36d8467c8327cfc6d7f30282e981ea6856e37f0956fce1fbb5d3) set the `Content-Encoding: gzip` header but sent an uncompressed body, causing malformed requests. The branch is removed so all bulk POST requests are sent uncompressed without a compression header. The `cmp_bytes` tracing field is also removed.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 6a20c33. 1 file reviewed, 1 issue evaluated, 0 issues filtered, 1 comment posted</summary>
>
> ### 🗂️ Filtered Issues
>
> </details>
>

### DIFF
--- a/crates/ffwd-output/src/elasticsearch/transport.rs
+++ b/crates/ffwd-output/src/elasticsearch/transport.rs
@@ -62,12 +62,7 @@ impl ElasticsearchSink {
         }
 
         tracing::Span::current().record("req_bytes", body_len as u64);
-        let req = if self.config.compress {
-            tracing::Span::current().record("cmp_bytes", body.len() as u64);
-            req.header("Content-Encoding", "gzip").body(body)
-        } else {
-            req.body(body)
-        };
+        let req = req.body(body);
 
         let t0 = std::time::Instant::now();
         let response = match req.send().await {


strawgate:
Closing: this fix would break working compression.

After PR #2580, the ES output compression works correctly on main:
1. `send.rs` compresses body with `GzEncoder` → passes compressed bytes to `do_send()`
2. `transport.rs::do_send()` adds `Content-Encoding: gzip` header when `compress=true`

Removing the header from `do_send()` while `send.rs` still compresses would send gzipped bytes without the header — the reverse of the original bug.

The `cmp_bytes` metric records `body.len()` which IS the compressed size (since `send.rs` already compressed it before calling `do_send()`). Redundant with `req_bytes` but not incorrect.

Issue #2564, opened 2026-04-24
bug: Journald subprocess restarts do not resume from a stable cursor


strawgate:
# Journald subprocess restarts do not resume from a stable cursor

## Summary
The subprocess backend restarts `journalctl` from static config, not from the last delivered cursor. That creates two user-visible variants: `since_now: false` replays retained history, while `since_now: true` skips entries that appear while the child is down.

## Slice
`journald input behavior`, `recovery`

## Target
- Reproduced on `origin/main`
- Commit: `27b42546d3179ef09e85b2f789ca72a7ca40c284`

## Dedup Check
- Existing hunt files checked: the six cluster candidates in `tmp/bug-hunt/2026-04-23-sweep-30x4/candidates/`
- Open issues searched: `journald subprocess restart since_now replay duplicate loss cursor`
- Closed issues / merged PRs searched: same keywords
- No direct matching issue/PR found

## Reproduction
1. Temporarily add two tests in [crates/logfwd-io/src/journald_input.rs](/Users/billeaston/.codex/worktrees/fastforward-bughunt-main/crates/logfwd-io/src/journald_input.rs) that run `JournaldInput` against a fake `journalctl` shell script:
   - one script prints the same retained record on every invocation, proving replay under `since_now: false`
   - one script prints a record only on the first invocation, proving the restarted `since_now: true` process can miss downtime data
2. Run:

```bash
CARGO_TARGET_DIR=/tmp/ff-bughunt-target cargo test -p logfwd-io subprocess_restart_with_since_now_ -- --nocapture
```

## Expected
After a subprocess crash or EOF, journald input should resume from the last delivered cursor instead of restarting from a fresh wall-clock or broad historical query.

## Actual
- `since_now: false`: duplicate record replay after restart
- `since_now: true`: no second record appears after restart even though the helper was reinvoked

## Evidence
- Restart loop respawns the same command from config: [crates/logfwd-io/src/journald_input.rs](/Users/billeaston/.codex/worktrees/fastforward-bughunt-main/crates/logfwd-io/src/journald_input.rs:817), [journald_input.rs](/Users/billeaston/.codex/worktrees/fastforward-bughunt-main/crates/logfwd-io/src/journald_input.rs:882)
- `since_now: true` maps to `--since=now`: [crates/logfwd-io/src/journald_input.rs](/Users/billeaston/.codex/worktrees/fastforward-bughunt-main/crates/logfwd-io/src/journald_input.rs:792)
[...]

Issue #2561, opened 2026-04-24
bug: TCP input drops live partial lines on clean shutdown


strawgate:
# TCP input drops live partial lines on clean shutdown

## Summary
`TcpInput` flushes pending partial lines when a connection dies, but clean shutdown calls the default no-op `poll_shutdown()`. Bytes already buffered in `client.pending` for still-open connections are discarded when the input is dropped.

## Slice
`tcp input behavior`, `shutdown/recovery`

## Target
- Reproduced on `origin/main`
- Commit: `27b42546d3179ef09e85b2f789ca72a7ca40c284`

## Dedup Check
- Existing hunt files checked: the six cluster candidates in `tmp/bug-hunt/2026-04-23-sweep-30x4/candidates/`
- Open issues searched: `checkpoint flush retry shutdown replay success tcp shutdown partial line`
- Closed issues / merged PRs searched: same keywords
- Related prior fix: issue `#804` fixed the disconnect path, not the clean-shutdown path

## Reproduction
1. Temporarily add a unit test in [crates/logfwd-io/src/tcp_input/tests/basic.rs](/Users/billeaston/.codex/worktrees/fastforward-bughunt-main/crates/logfwd-io/src/tcp_input/tests/basic.rs) that:
   - opens a TCP input,
   - sends `partial-without-newline`,
   - polls until the bytes are sitting in `input.clients[0].pending`,
   - calls `input.poll_shutdown()`.
2. Run:

```bash
CARGO_TARGET_DIR=/tmp/ff-bughunt-target cargo test -p logfwd-io verification_tcp_shutdown_ignores_live_partial_line -- --nocapture
```

## Expected
Clean shutdown should flush already-read partial data and/or emit EOF for the live source so the remainder is not lost.

## Actual
`poll_shutdown()` returns no events while the partial line is still buffered in `client.pending`.

## Evidence
- Default shutdown hook is a no-op: [crates/logfwd-io/src/input.rs](/Users/billeaston/.codex/worktrees/fastforward-bughunt-main/crates/logfwd-io/src/input.rs:395)
- TCP only flushes partial tails for dead connections: [crates/logfwd-io/src/tcp_input/input_source.rs](/Users/billeaston/.codex/worktrees/fastforward-bughunt-main/crates/logfwd-io/src/tcp_input/input_source.rs:239)
- Verification command output showed:
  - `test tcp_input::tests::verification_tcp_shutdown_ignores_live_partial_line ... ok`

## Severity
[...]

PR #2706, opened 2026-04-27
fix: flush pending TCP partial lines on shutdown and resume journald from last cursor

strawgate:
## Summary

Two related fixes for input handling:

### Bug #2561: TCP input drops live partial lines on clean shutdown
When `TcpInput::poll_shutdown()` was called on a live connection with pending partial lines (no newline), those bytes were dropped instead of being flushed.

**Fix:** Added `poll_shutdown()` method to `TcpInput` that:
- Flushes pending partial lines with a synthetic newline before closing
- Emits `EndOfFile` for each client

### Bug #2564: Journald subprocess restarts do not resume from a stable cursor
When the `journalctl` subprocess crashed and restarted, it would restart from a static `--since=` timestamp instead of the last delivered cursor. This caused:
- `since_now: false`: duplicate replay of history after restart
- `since_now: true`: missing entries that appeared during downtime

**Fix:** Added cursor tracking to the subprocess backend:
- Extract and save `__CURSOR` from each journal entry after successful delivery
- On restart, pass `--cursor=LAST_CURSOR` to resume from the correct position
- Skip the first entry on restart if it matches the cursor we restarted from (to avoid duplicate)

## Changes

| File | Change |
|------|--------|
| `crates/ffwd-io/src/tcp_input/input_source.rs` | Added `poll_shutdown()` implementation |
| `crates/ffwd-io/src/tcp_input/tests/basic.rs` | Added test for shutdown flush behavior |
| `crates/ffwd-io/src/journald_input.rs` | Added cursor tracking and restart handling |

## Validation

- `cargo test -p ffwd-io`: 35 passed, 12 ignored
- `cargo test -p ffwd-runtime`: 248 passed
- `cargo clippy -p ffwd-io -p ffwd-runtime`: clean
- `cargo fmt --check`: clean

## Design Notes

- The journald cursor is saved AFTER successful send to avoid duplicates on restart
- The `restart_cursor` parameter to `read_export_entries` tracks whether we're resuming and skips the duplicate entry
- The subprocess backend now uses `--cursor=` when a cursor is available, falling back to `--since=` only on first start

> [!NOTE]
> ### Flush pending TCP partial lines on shutdown and resume journald subprocess from last cursor
> - Adds `poll_shutdown` to [`InputSource`](https://github.com/strawgate/fastforward/pull/2706/files#diff-f8e72ab4a4e6e575ba665bde31164ff37d2c0b96c5cead716f98179144382d54) that emits a final `Data` event for any buffered partial line followed by an `EndOfFile` event per client, ensuring no data is lost on shutdown.
> - The journald subprocess backend in [`journald_input.rs`](https://github.com/strawgate/fastforward/pull/2706/files#diff-6940a2fb0fe8ebffe545378af75bcb4fc6132a5916b581d219f5fb7241aec757) now tracks the last seen `__CURSOR` across restarts, passing it to `journalctl` via `--cursor` so reads resume without replay or gaps.
> - `build_command` now accepts an optional cursor; when present it adds `--cursor=<value>` and suppresses any `--since` flag.
> - `RetryConfig.max_attempts` in [`shared.rs`](https://github.com/strawgate/fastforward/pull/2706/files#diff-ae33b990c0b4f0f73c8cd98eec648027f1ffc24afa380182a180aa18d3301f62) is changed from `Option<i32>` to `Option<u32>`, rejecting negative values at deserialization time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 921d27d.</sup>
> 
>

### DIFF
--- a/crates/ffwd-config/src/shared.rs
+++ b/crates/ffwd-config/src/shared.rs
@@ -60,7 +60,7 @@ pub struct TlsServerConfig {
 #[serde(deny_unknown_fields)]
 pub struct RetryConfig {
     #[serde(default, deserialize_with = "deserialize_option_from_string_or_value")]
-    pub max_attempts: Option<i32>,
+    pub max_attempts: Option<u32>,
     #[serde(default, deserialize_with = "deserialize_option_from_string_or_value")]
     pub initial_backoff_secs: Option<PositiveSecs>,
     #[serde(default, deserialize_with = "deserialize_option_from_string_or_value")]
--- a/crates/ffwd-config/tests/validation_gaps.rs
+++ b/crates/ffwd-config/tests/validation_gaps.rs
@@ -1425,3 +1425,75 @@ pipelines:
     Config::load_str(yaml)
         .expect("push path with trailing slash should be accepted (normalized by output)");
 }
+
+#[test]
+fn issue_2584_retry_max_attempts_rejects_negative_values() {
+    for max_attempts in [-1, -100] {
+        let yaml = format!(
+            r#"
+pipelines:
+  test:
+    inputs:
+      - type: generator
+        generator:
+          profile: record
+    outputs:
+      - type: elasticsearch
+        endpoint: http://localhost:9200
+        retry:
+          max_attempts: {max_attempts}
+"#,
+        );
+        let result = Config::load_str(&yaml);
+        assert!(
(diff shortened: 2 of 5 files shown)


strawgate:
TCP partial shutdown fix is merged

